PR #1, opened 2023-04-12
Add ESP32-S3 example

bjoernQ:
Add ESP32-S3 example and some cleanup.
This needs a recent commit of `esp-wifi`


JurajSadel (approved):
LGTM, I tested all examples and they are working correctly, only one note, be sure you are using `--release`.

PR #5, opened 2023-04-26
Add CI

bjoernQ:
This adds CI.

I tried something new here: Using a `justfile` to run the actual checks and use as-little-as-possible of GitHub Actions

- good: this makes it really easy to do the same checks locally (just use `just`)
- downside: it's harder to see at a glance what exactly failed
- downside: if the job fails, rerunning means rerunning everything

I'd like to experiment with this a bit since the first point is really something I'd like to have - maybe there are even solutions for the other two points

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,43 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - main
+  push:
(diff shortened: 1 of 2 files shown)


jessebraham (approved):
I don't know anything about `just` personally so can't comment on the `justfile`. CI config looks fine otherwise though.

PR #6, opened 2023-04-27
Ensure that certificates are null terminated

AnthonyGrondin:
This is a quality of life PR.

Mbedtls requires that certificates and passwords passed to it are null terminated `\0`, like C strings.  

This new helper checks if the certificates are null-terminated, and if not, it adds the nullish terminator.

This allow to include certificates using `include_str!()` without having to append `\0` ourselves, since the library takes care of it.

### DIFF
--- a/esp-mbedtls/src/lib.rs
+++ b/esp-mbedtls/src/lib.rs
@@ -15,6 +15,8 @@ use embedded_io::Io;
 use esp_mbedtls_sys::bindings::*;
 use esp_mbedtls_sys::c_types::*;
 
+use crate::compat::ensure_null_terminated;
+
 // these will come from esp-wifi (i.e. this can only be used together with esp-wifi)
 extern "C" {
     fn free(ptr: *const u8);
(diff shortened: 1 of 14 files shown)


AnthonyGrondin:
Currently having an Error when using a private key that isn't null terminated. Trying to figure that one out.

bjoernQ:
For certs at least maybe taking a byte slice instead of `&str` is a good thing (plus checking the null termination at least in debug mode) 

Adding the `\0` ourselves would be very convenient but we need to copy all the data to RAM then.

For passwords we can do the same or do the null-termination for convenience (since they are not that huge) - we just should make sure that we wipe out the copied bytes of the password after usage

AnthonyGrondin:
Since it would be easier to take a reference to a slice, and let the user manage how they store their certificates,
I've moved over most of the code from https://github.com/esp-rs/esp-idf-svc/blob/master/src/tls.rs, but we return an error if the pem doesn't end with a null byte, instead of panicking.

I've tested with both der and pem, and it seems to work. I'm waiting for review before updating the other examples.

esp-mbedtls/src/lib.rs
```diff
@@ -87,10 +88,83 @@ pub fn set_debug(level: u32) {
...
+/// const CERTIFICATE: &[u8] = include_bytes!("certificate.der");
+/// let cert = X509::der(CERTIFICATE);
+/// ```
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub struct X509<'a> {
+    /// The certificate in bytes
+    bytes: &'a [u8],
+
+    /// The length of the certificate
+    len: usize,
```
bjoernQ, line 111: maybe I missed something but the `bytes` slice already knows it's length so we don't need to store the length here

esp-mbedtls/src/lib.rs
```diff
@@ -87,10 +88,83 @@ pub fn set_debug(level: u32) {
...
+    }
+
+    /// Returns a pointer to the data for parsing
+    pub(crate) fn as_ptr(&self) -> *const c_uchar {
+        self.data().as_ptr().cast()
+    }
+
+    /// Gets the first null byte in a slice
+    fn get_null(bytes: &[u8]) -> Option<usize> {
+        bytes.iter().position(|&byte| byte == 0)
```
bjoernQ, line 160: wouldn't it be good enough to check the last byte?

esp-mbedtls/src/lib.rs
AnthonyGrondin: This is for a specific use case where bytes would be a `[0u8; T]`, with extra 0s at the end, the certificate not taking the full size of it.

esp-mbedtls/src/lib.rs
AnthonyGrondin: Same as https://github.com/esp-rs/esp-mbedtls/pull/6#discussion_r1183398663

bjoernQ (approved):
LGTM

bjoernQ:
Thanks for the reminder - I somehow forgot about this

PR #7, opened 2023-06-15
Add #[derive] attributes to certificates

AnthonyGrondin:
Add useful #[derive] attributes to certificates, like copy, which allows to reuse the certificates after creating a new session.


bjoernQ (approved):
LGTM

Thanks for your contribution

PR #8, opened 2023-06-22
Update esp-wifi and esp-hal

AnthonyGrondin:
Update `esp-wifi` and `esp-hal` to the latest version.

I've currently only done esp32s3, because it's the only chip that I have on hand. 

I'm getting multiple crashes, for many different reasons, like illegal instructions, or index out of bounds.. I'm also getting this kind of error on another project using this library, so it feels like it's specific to esp-wifi and / or esp-mbedtls. I can't figure this one out.

Tested on both `1.69.0` and `1.70.0`.

### DIFF
--- a/examples-esp32c3/Cargo.toml
+++ b/examples-esp32c3/Cargo.toml
@@ -7,6 +7,16 @@ license = "MIT OR Apache-2.0"
 
 [profile.release]
 debug = true
+lto = false
+
+[profile.release.package.esp-wifi]
+opt-level = 3
+
+[profile.dev]
+lto = false
+
+[profile.dev.package.esp-wifi]
+opt-level = 3
 
 [dependencies]
 hal = { package = "esp32c3-hal", version = "0.9.0" }
--- a/examples-esp32c3/Cargo.toml
+++ b/examples-esp32c3/Cargo.toml
@@ -24,7 +24,7 @@ esp-backtrace = { version = "0.6.0", features = ["esp32c3", "panic-handler", "pr
 esp-println       = { version = "0.4.0", features = ["esp32c3","log"] }
 
 embassy-time = { version = "0.1.0", features = ["nightly"], optional = true }
-embassy-executor = { version = "0.2.0", package = "embassy-executor", features = ["nightly", "executor-thread", "integrated-timers", "arch-xtensa"], optional = true }
+embassy-executor = { version = "0.2.0", package = "embassy-executor", features = ["nightly", "executor-thread", "integrated-timers", "arch-riscv32"], optional = true }
 embassy-net = { version = "0.1.0", features = ["nightly", "tcp", "udp", "dhcpv4", "medium-ethernet"], optional = true }
 
 esp-wifi = { git = "https://github.com/esp-rs/esp-wifi.git", features = ["big-heap", "phy-enable-usb", "esp32c3", "embedded-svc", "wifi"], rev = "68dc11bbb2c0efa29c4acbbf134d6f142441065e" }
--- a/examples-esp32c3/Cargo.toml
+++ b/examples-esp32c3/Cargo.toml
@@ -19,16 +19,16 @@ lto = false
 opt-level = 3
 
 [dependencies]
-hal = { package = "esp32c3-hal", version = "0.9.0" }
+hal = { package = "esp32c3-hal", version = "0.10.0" }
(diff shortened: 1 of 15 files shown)


bjoernQ:
I can reproduce the `IllegalInstruction` with the `sync` example - results in this (1.70, 1.69 is similar with slightly different addresses)

```
Exception occured 'IllegalInstruction'
Context
PC=0x4037dacf       PS=0x00060011
0x4037dacf - hal_pwr_interrupt_clr_event
    at ??:??
0x00060011 - PS_WOE
    at ??:??
A0=0x8037da21       A1=0x3fcd9e70       A2=0x00000000       A3=0x00000001       A4=0x60038004
0x8037da21 - _rtc_fast_bss_start
    at ??:??
0x3fcd9e70 - _heap_end
...
```

With 1.68 it works however.

Given the PC is in IRAM I suspect something increased stack usage (plus maybe 1.69 and 1.70 reserve more stack somehow)

Not sure if this is caused by esp-hal or esp-wifi (probably it's esp-wifi). I had a look at the diff but there is nothing really obvious wasting stack.

Also https://github.com/esp-rs/esp-wifi/issues/196 suggests it's not an Xtensa-only thing.

However, I cannot reproduce it with the example here for ESP32-C3

bjoernQ:
Update: Most probably not a problem with the stack BUT a linker script problem

```
4037dacf:	f01d      	retw.n

3fc8dad0 A _data_start
3fc8dad0 d _ZN8esp_wifi4wifi10os_adapter14FAKE_SPIN_LOCK17h713b30e6a3223c01E
```

bjoernQ:
I can confirm it's a linker script issue. Needs some more testing and then I can hopefully create a PR for `esp-hal` tomorrow.

AnthonyGrondin:
I've added the examples for esp32 and esp32c3. I haven't tested if they run, as I don't have those chips on hand, but I've tested for esp32s3 and everything works fine, except for `sync_client` which seems to hang for me, at `Making HTTP request`

bjoernQ:
The problem on ESP32-S3 should be fixed in https://github.com/esp-rs/esp-hal/pull/616 - maybe you want to check it by patching the dependency

bjoernQ:
Seems like it's working for all the targets now with a patched HAL. We plan to have new HAL releases soon (and update esp-wifi as soon as the HALs are released)

AnthonyGrondin:
Great! Thanks for taking care of it.
I'll wait until `esp-hal` and `esp-wifi` get updated, then update the changes in this PR.

AnthonyGrondin:
Needs testing on `esp32` and `esp32c3`, but I've tested `esp32s3` and everything works.

This gives me errors when I compile without `--release`

examples-esp32c3/Cargo.toml
```diff
@@ -7,19 +7,29 @@ license = "MIT OR Apache-2.0"
...
 [dependencies]
-hal = { package = "esp32c3-hal", version = "0.8.0" }
+hal = { package = "esp32c3-hal", version = "0.10.0" }
 esp-backtrace = { version = "0.6.0", features = ["esp32c3", "panic-handler", "print-uart", "exception-handler"] }
 esp-println       = { version = "0.4.0", features = ["esp32c3","log"] }
 
 embassy-time = { version = "0.1.0", features = ["nightly"], optional = true }
-embassy-executor  = { package = "embassy-executor", git = "https://github.com/embassy-rs/embassy/", rev = "cd9a65b", features = ["nightly", "integrated-timers"], optional = true }
-embassy-net = { git = "https://github.com/embassy-rs/embassy", rev = "26474ce6eb759e5add1c137f3417845e0797df3a", features = ["nightly", "tcp", "udp", "dhcpv4", "medium-ethernet"], optional = true }
+embassy-executor = { version = "0.2.0", package = "embassy-executor", features = ["nightly", "executor-thread", "integrated-timers", "arch-xtensa"], optional = true }
```
bjoernQ, line 27: Probably copy+paste: We need `"arch-riscv32"` instead of `"arch-xtensa"`

AnthonyGrondin: Yeah, it slipped in with a copy + paste. Thanks for pointing it out.

bjoernQ (approved):
LGTM

bjoernQ:
Compiles fine now for ESP32-C3. I had some problems with the ESP32-C3 `async_client` example since it seems that something (probably the WiFi drivers) now allocate more memory and I had to remove the last certificate in `CERT` to make it work.

I will merge this anyways to not block you from making progress in #9

PR #18, opened 2023-09-15
Don't return from `read()` if there's no data

### LINKED ISSUE #16: `read()` returns `Ok(0)` if there's no data
This is an issue similar as https://github.com/esp-rs/esp-wifi/pull/216 where `Ok(0)` is returned from `read()` if there's no data.

This causes issues when trying to use the `read()` impl from inside other crates, since it might interpret `Ok(0)` as an EOF. This is what happens initially during a request handshake.

This forces the use of something like:
```rust
loop {
    match tls.read(&mut temp_buffer).await {
        Ok(0) => continue,
        Ok(_) => break,
        Err(TlsError::Eof) => {
            log::error!("EOF");
            break;
        }
        Err(e) => {
            log::error!("Read Error {:?}", e);
            break;
        }
    };
}
```

and then use this temporary buffer for the other crates.

For example, trying to use the current `read()` impl inside https://github.com/drogue-iot/reqwless/blob/f23567d8d178dd837332b503fe983ba4931d8b40/src/response.rs#L47-L58 will result in no data being read.

AnthonyGrondin:
Fix #16

### DIFF
--- a/esp-mbedtls/src/lib.rs
+++ b/esp-mbedtls/src/lib.rs
@@ -534,21 +534,15 @@ where
     T: Read + Write,
 {
     fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
-        let res = self.session.internal_read(buf);
-        if res <= 0 {
-            if res == MBEDTLS_ERR_SSL_WANT_READ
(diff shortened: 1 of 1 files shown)


bjoernQ (approved):
LGTM - Thanks

PR #20, opened 2023-11-04
feat(bignum): Add initial bignum hardware acceleration

AnthonyGrondin:
## Description
This is still work in progress.

I've tried to port over most of the code from https://github.com/espressif/esp-idf/blob/b4268c874a4cf8fcf7c0c4153cffb76ad2ddda4e/components/mbedtls/port/bignum/esp_bignum.c

Using espressif's port of mbedtls, we can provide custom implementation for `mpi` operations, using the RSA hardware accelerator.

I've only made it work for `esp32s3`  with the examples `crypto_self_test` and `sync_server` so far. 
I will fix the compilation for async and other targets and examples when I get this working.

### What works:
 - The self tests seem to work, 

### What doesn't work:
- When I use the sync_server example, the RSA verification fails, which I suspect might be something happening with bigger keys, since the self tests only use less than 2048 long keys, while the server uses longer keys.

- `mbedtls_mpi_mult_mpi_failover_mod_mult()` used to use mod multi for bigger keys seems to be broken too, as the assert fails. I've commented it out, to use the overlong, even if it takes more time, to isolate the issue mentioned above.

### Building and running
Libraries need to be need to be rebuilt, and the submodule refreshed on existing clones, since I've changed it to use espressif's custom fork.

```bash
cargo +stable xtask compile
cargo +stable xtask bindings
```


AnthonyGrondin:
I've switched over to try to implement the modular exponentiation, and I've managed to get it work on the `esp32s3` for all self_tests, and `sync_server`. I did some limited benchmarks and it's slightly faster than using the software implementation. There might still be ways to tweak it and make it faster.  

There are still safety issues, because it seems to fail when using it in async and sometimes the verification fails.

I still have to figure out how to use Generics for RSA to reduce code duplication, as we need to provide a generic from a non-const value at runtime (max operand size). The current architecture of the RSA module makes it tricky.

PR #25, opened 2024-02-12
revert: Changes Session back to own the socket

AnthonyGrondin:
This was done as a fix for the server, but it seems that letting T and simply passing &mut T fixes the original issue this PR was intended to fix.

Other libraries like `embedded_tls` take ownership of the socket. This allows to move the session around without having to keep a reference to the socket alive.

I've ran the justfile and everything compiles. This unblocks my fork of reqwless that attempts to use `esp-mbedtls` as a TLS wrapper (_coming soon_).

Ref: c094cfc

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,7 +5,6 @@ rustflags = [
   "-C", "link-arg=-Tlinkall.x",
   "-C", "link-arg=-nostartfiles",
   "-C", "link-arg=-Trom_functions.x",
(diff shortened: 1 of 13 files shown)


bjoernQ:
I think the build fails because of the changes around atomics (not related to this PR). We probably need to update the HAL dependencies and remove the atomic-emulation configs from `.cargo/config.toml`

bjoernQ (approved):
LGTM - Thanks for taking care of updating the dependencies!

PR #27, opened 2024-02-26
Re-export Rsa since it is a parameter of Session::new()

AnthonyGrondin:
This is a quality of life improvement, as it allows a library to wrap esp-mbedtls without having to re-import the hal for the Rsa struct.


bjoernQ (approved):
LGTM

PR #28, opened 2024-03-16
Update to esp-hal 0.16.1

AnthonyGrondin:
Just a simple update to `esp-hal 0.16.1` and use cargo aliases in the justfile.


bjoernQ (approved):
LGTM - I think the CI problem is a problem of the current nightly also encountered here: https://github.com/esp-rs/esp-hal/pull/1301

I guess we should merge this anyways?

AnthonyGrondin:
Either way is fine with me. `justfile` runs fine without error on my side.

PR #30, opened 2024-07-10
Update to esp-hal 0.18.0

AnthonyGrondin:
- I've updated `esp-hal` to the latest current release. 
- I've locked the nightly version to `nightly-2024-06-12` until the fix in https://github.com/embassy-rs/embassy/issues/3078 gets released
- I've reduced `heap_size` in config to be able to build without overflowing segments.

:warning: Breaking Change :warning: 
- The `rsa` field of `Session::new()` now take the peripheral directly, instead of an instance of the driver, so that we can instantiate the driver ourselves, in blocking mode. This now also allow creating a session in a trait impl that immutably borrow `&self` without triggering the borrow checker, by declaring a field with `RefCell<Option<esp_hal::peripherals::RSA>>`, and calling `Session::new()` with the following for the rsa field:
```rust
if let Ok(ref mut rsa) = self.rsa.try_borrow_mut() {
    rsa.as_mut()
} else {
    None
},
```

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -47,12 +47,12 @@ xtask = "run --manifest-path ./xtask/Cargo.toml --"
 # Alias' for quickly building for different chips or running examples
 # By default we enable
 #   - `default` HAL features to set up basic chip specific settings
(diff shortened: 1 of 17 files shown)


bjoernQ (approved):
Thanks! Looks good to me

PR #31, opened 2024-07-11
Use a builder pattern for enabling hardware accelerated RSA

AnthonyGrondin:
Built on top of #30.

This PR refactors the way we pass the RSA peripheral to a Session, using the builder pattern.
This also adds documentation detailing the usage of hardware accelerated RSA.

Whereas the older implementation is on a per session basis, it comes with the issue that only a single session can use the peripheral at the same time, even if we hold RSA statically. It would also be overwritten for every session, if one session is initialized with `None`.

This implementation is closer to the actual usage we do with the peripheral, where it is passed through the FFI with a static, hence it will be used for every session. The RSA is safely dropped of usage when the session using the peripheral is dropped.

It also improves simplicity, because with the change in #30, using `None` for `RSA` would require specifying the type with `None::<esp_hal::peripherals::RSA>`.

### DIFF
--- a/esp-mbedtls/src/lib.rs
+++ b/esp-mbedtls/src/lib.rs
@@ -29,6 +29,12 @@ pub use esp_mbedtls_sys::bindings::{
 use esp_mbedtls_sys::c_types::*;
 
 /// Hold the RSA peripheral for cryptographic operations.
+///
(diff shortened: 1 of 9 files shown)


bjoernQ (approved):
Thanks - LGTM

PR #32, opened 2024-07-12
Use `mbedtls_x509_crt_parse_der_nocopy()` when certificate is in DER format

### LINKED ISSUE #17: Do not make a copy of the X509 when parsing it to save RAM.
We currently already own `Certificates<'a>` for the lifetime of the session, so this should be a drop-in replacement. This would allow to re-use the same certs in multiple sessions, or with the client and the server at the same time without making unnecessary copies.

I'll paste the [doc](https://mbed-tls.readthedocs.io/en/latest/kb/how-to/reduce-polarssl-memory-and-storage-footprint/#x-509) here:

> # X.509[:link:](https://mbed-tls.readthedocs.io/en/latest/kb/how-to/reduce-polarssl-memory-and-storage-footprint/#x-509)
> ## Parsing X.509 certificates without copying the raw certificate data[:link:](https://mbed-tls.readthedocs.io/en/latest/kb/how-to/reduce-polarssl-memory-and-storage-footprint/#parsing-x-509-certificates-without-copying-the-raw-certificate-data)
> 
> The X.509 CRT parsing APIs `mbedtls_x509_crt_parse()` and `mbedtls_x509_crt_parse_der()` create an internal copy of the raw certificate data passed to them. While this allows you to free or reuse the input buffer, it means the raw certificate data will be twice in memory at some point.
> 
> To avoid that, the following API can be used to set up an X.509 certificate structure without making a copy of the input buffer:
> 
> ```c
> int mbedtls_x509_crt_parse_der_nocopy( mbedtls_x509_crt *chain,
>                                        const unsigned char *buf,
>                                        size_t buflen );
> ```
> The only difference between `mbedtls_x509_crt_parse_der_nocopy()` and `mbedtls_x509_crt_parse_der()` is that the buffer passed to `mbedtls_x509_crt_parse_der_nocopy()` holding the raw DER-encoded certificate must stay unmodified for the lifetime of the established X.509 certificate context. See the [documentation](https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/x509_crt.h) for more information.
> 
> *Example*: If your own certificate and/or the trusted CA certificates are hardcoded in ROM, you may use `mbedtls_x509_parse_der_nocopy()` to create X.509 certificate contexts from them without an additional copy in RAM.

AnthonyGrondin:
Built on top of #30 

Fix #17

I've measured the free heap size while using `async_server.rs` to compare between DER and PEM in local experiments, and this saves about **1KB** of heap memory taken by a session.

### DIFF
--- a/esp-mbedtls/src/lib.rs
+++ b/esp-mbedtls/src/lib.rs
@@ -106,6 +106,13 @@ pub fn set_debug(level: u32) {
     }
 }
 
+/// Format type for [X509]
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+enum CertificateFormat {
+    PEM,
+    DER,
+}
+
 /// Holds a X509 certificate
 ///
 /// # Examples
@@ -121,7 +128,10 @@ pub fn set_debug(level: u32) {
 /// let cert = X509::der(CERTIFICATE);
 /// ```
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
-pub struct X509<'a>(&'a [u8]);
+pub struct X509<'a> {
+    bytes: &'a [u8],
+    format: CertificateFormat,
+}
 
 impl<'a> X509<'a> {
     /// Reads certificate in pem format from bytes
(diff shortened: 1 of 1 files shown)


bjoernQ (approved):
Thanks

PR #38, opened 2024-07-18
Implement `edge_nal::TcpAccept` trait from `edge-net`

AnthonyGrondin:
This PR provides an implementation of [edge_nal::TcpAccept](https://github.com/ivmarkov/edge-net/blob/640a4f263986d6d26e2508cdb2f43c8ee5a7bd66/edge-nal/src/stack/tcp.rs#L68-L83) to use `esp-mbedtls` with `edge-net`.

As for now, the trait is simply a wrapper over the [edge_nal::TcpAccept implementation](https://github.com/ivmarkov/edge-net/blob/640a4f263986d6d26e2508cdb2f43c8ee5a7bd66/edge-nal-embassy/src/tcp.rs#L63-L91) from `edge-nal-embassy`. 

I also provided an example that uses `edge-http` in server mode for testing, and ensuring compatibility.

Further down the line, other traits will also get their implementation, in subsequent PRs, as I (or any contributor) come around a use case for them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,9 @@ static_cell = { version = "=1.2", features = ["nightly"] }
 
 esp-mbedtls = { path = "./esp-mbedtls" }
 
+edge-http = { package = "edge-http", git = "https://github.com/ivmarkov/edge-net/", optional = true }
+edge-nal-embassy = { package = "edge-nal-embassy", git = "https://github.com/ivmarkov/edge-net/", optional = true }
+
 [[example]]
 name = "crypto_self_test"
 required-features = ["esp-wifi/wifi-logs"]
@@ -86,6 +89,10 @@ required-features = ["async"]
 [[example]]
 name = "async_server_mTLS"
 required-features = ["async"]
+
+[[example]]
+name = "edge_server"
+required-features = ["async", "esp-hal-embassy", "edge-nal-embassy", "edge-http", "esp-mbedtls/edge-nal"]
 
 [features]
 esp32 = [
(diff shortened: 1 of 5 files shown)


AnthonyGrondin:
I'll write this here to document the behavior, but there is currently an issue with how `esp-wifi`, `esp-mbedtls`, `edge-net` act together.

Specifically, if a client sends a request with the header `"Connection": "Close"`, the server implementation of `edge-net` requests a `read()` from `esp-mbedtls` (It won't if the body type is something else). Since the client has sent all its data, `async_internal_read()` will hang on reading data from socket. This is due to `esp-wifi`'s read impl that blocks on read until there is data.

ivmarkov:
> I'll write this here to document the behavior, but there is currently an issue with how `esp-wifi`, `esp-mbedtls`, `edge-net` act together.
> 
> Specifically, if a client sends a request with the header `"Connection": "Close"`, the server implementation of `edge-net` requests a `read()` from `esp-mbedtls` (It won't if the body type is something else). Since the client has sent all its data, `async_internal_read()` will hang on reading data from socket. This is due to `esp-wifi`'s read impl that blocks on read until there is data.

I'm not sure I follow 100% but regardless, are you using the `master` version of `edge-http`?
[There is a fix in `master` I did recently](https://github.com/ivmarkov/edge-net/commit/6b74a36d25e41146388f64fc93f32b7e4dff1336) which avoids a similar (yet, the opposite) problem, where if there are multiple requests scheduled over a single TCP connection, the processing of the first request might read a bit into the second request.

bjoernQ (approved):
Thanks

PR #45, opened 2024-08-30
Update esp-hal to `0.20.1`. Decompose RX, TX buffer sizes. Fix examples.

AnthonyGrondin:
This PR provides the following changes:

- Update esp-hal to `0.20.1`.
- The RX and TX buffers can now be set to use independent sizes.
- Re-order rx, tx according to discussed API guidelines.
- Fix `edge_server.rs` example by using a buffer pool like in `edge_nal_embassy`.
- Fix async server examples who called `make_static!()` more than once.
- Change xtask aliases to always build in release mode
- Move `edge_nal_compat` module to its own file to declutter `lib.rs`

Closes #44

PR #46, opened 2024-08-30
Feat/hw accelerated sha updated

AnthonyGrondin:
Built on top of #45 
Supersedes #19 

~Current status: Draft until the following topics have been resolved.~

- [x] Figure out why the binary size is larger when using HW sha.
  (Solved: Now binary size is smaller when using HW sha)
- [x] Benchmark a clients and servers to see if there's a significant performance increase / decrease.
  (Solved: Server example is slightly faster and self-tests are marginally faster)
- [x] Once binary size and speed figured out, decide if this PR is a good tradeoff overall.
  (Solved: The numbers speak for themselves)
- [x] Figure out the case for `esp32`. Mbedtls combines both SHA224 and SHA256 in the same function calls, but `esp32`  only support SHA256.
  (Solved: For now, `esp32` still uses SW sha for SHA224 and SHA256)

/cc @bugadani @Dominaezzz
Pinging you since this has been discussed in the matrix channel.


Dominaezzz:
Regarding benchmarks, I found [this](https://github.com/espressif/arduino-esp32/issues/6819#issuecomment-1140884110) earlier today.

AnthonyGrondin:
Oh wow, good find. I would've never thought about this.

AnthonyGrondin:
Now that esp_hal had a new release, I rebased on top of the current HEAD and removed the patches.
I also fixed a few missing todos and a memory leak bug.

I've a few metrics side by side on an esp32s3:

**Self-tests:**
| Hash Algorithm | Software (cycles) | Hardware (cycles) | Hardware Faster (x times) |
|----------------|-------------------|-------------------|---------------------------|
| SHA-1          |      3,356,883     |       654,913     |           5.12            |
| SHA-224        |      8,219,260     |       653,337     |           12.58           |
| SHA-256        |      8,206,023     |       650,996     |           12.61           |
| SHA-384        |     13,575,266     |       587,879     |           23.09           |
| SHA-512        |     13,558,071     |       584,100     |           23.21           |

**edge_server example using SW SHA:**
```
App/part. size:    999,344/1,048,576 bytes, 95.30%
Benchmark 1: curl -k https://192.168.69.166
  Time (mean ± σ):      2.436 s ±  0.023 s    [User: 0.003 s, System: 0.004 s]
  Range (min … max):    2.408 s …  2.470 s    10 runs
```

**edge_server example using HW SHA:**
```
App/part. size:    983,040/1,048,576 bytes, 93.75%
Benchmark 1: curl -k https://192.168.69.166
  Time (mean ± σ):      2.406 s ±  0.023 s    [User: 0.003 s, System: 0.004 s]
  Range (min … max):    2.375 s …  2.464 s    10 runs
```

esp-mbedtls/src/lib.rs
```diff
@@ -471,6 +489,10 @@ impl<T> Session<T> {
         min_version: TlsVersion,
         certificates: Certificates,
     ) -> Result<Self, TlsError> {
+        // TODO: Take peripheral from user
+        let sha = Sha::new(unsafe { SHA::steal() });
+        critical_section::with(|cs| SHARED_SHA.borrow_ref_mut(cs).replace(sha));
+
```
AnthonyGrondin, line 495: Still need to find a user-friendly way to take hold of the SHA peripheral.

AnthonyGrondin:
I'll put this in Draft until we can get the compilation to work with `edge_server` (edge-net). This is currently due to a lifetime issue, where we cannot statically borrow in a loop.

The lifetime issue is currently being fixed in:
 - https://github.com/embassy-rs/embassy/pull/3329
 - https://github.com/ivmarkov/edge-net/pull/35

esp-mbedtls/Cargo.toml
```diff
@@ -15,6 +15,7 @@ cfg-if = "1.0.0"
 edge-nal = { version = "0.3.0", optional = true }
 edge-nal-embassy = { version = "0.3.0", optional = true }
 embassy-net = { version = "0.4", features = [ "tcp", "medium-ethernet"], optional = true }
+critical-section = "1.0.1"
```
bjoernQ, line 18: I think we are using "1.1.3" everywhere else currently

Cargo.toml
```diff
@@ -32,7 +32,7 @@ esp-println = { version = "0.12.0", features = ["log"] }
 esp-hal-embassy = { version = "0.4.0", optional = true }
 
 embassy-time = { version = "0.3.0", optional = true }
-embassy-executor = { version = "0.6.0", package = "embassy-executor", features = [
+embassy-executor = { version = "=0.6.0", package = "embassy-executor", features = [
```
bjoernQ, line 35: can't we just use 0.6.1 and not patch the crate?

bjoernQ:
For whatever reason I get compile errors for the C3 and async (but not S3?) - using embassy-executor 0.6.1 (and not patching any embassy crate) makes it work for me. Wonder why it works in CI 🤔

AnthonyGrondin:
This is due to changes in the nightly wakers, where embassy-executor `0.6.1` uses the new API, whereas `0.6.0` uses the existing API, that builds up to `nightly-2024-09-06` I believe.

Once 1.83 is released, we can use embassy-executor `0.6.1` and force the MSRV to 1.83. That being said, this is all for examples, so it shouldn't affect the inner workings of the crate.

bjoernQ:
> This is due to changes in the nightly wakers, where embassy-executor `0.6.1` uses the new API, whereas `0.6.0` uses the existing API, that builds up to `nightly-2024-09-06` I believe.
> 
> Once 1.83 is released, we can use embassy-executor `0.6.1` and force the MSRV to 1.83. That being said, this is all for examples, so it shouldn't affect the inner workings of the crate.

Ah yeah - thanks for the explanation (which also explains why it fails locally for me using a recent nightly)

bjoernQ (approved):
Thanks

PR #48, opened 2024-09-09
fix(async_internal_read): Ensure that all pending data has been processed before returning `Ok(0)`

AnthonyGrondin:
This fixes a weird issue I was having while making TLS connections as a client, where `Ok(0)` could be returned early from `async_internal_read()`, while there are still records to be processed from the ssl context.


bjoernQ (approved):
Thanks for your continued effort you put into this

PR #50, opened 2024-09-26
Pull `edge-net` crates from crates.io instead of git dependency.

AnthonyGrondin:
Use the released version of `edge-nal` and `edge-nal-embassy`.


bjoernQ (approved):
Thanks!

PR #51, opened 2024-10-03
upgrade to esp-wifi 0.9.1

yanshay:
removed duplicate functions
increased heap memory
updated examples cargo.toml


yanshay:
Can this be merged or should I change anything?

It's important since with `esp-wifi` 0.8.0 I encountered issues with getting IP, which I don't see with 0.9.1, and also there's a fix coming in the next versions to resolve some other issues getting IP, so need to get closer to vNext.

bjoernQ:
Thanks!

`StrBuf::from` in `compat/mod.rs` seems to be unused now and can get removed.

The examples should be changed to use `esp-alloc` as a global allocator like here: https://github.com/esp-rs/esp-hal/blob/c7a236884589f33fe59b302154dfaa18c9af110c/examples/src/bin/wifi_dhcp.rs#L52

yanshay:
> `StrBuf::from` in `compat/mod.rs` seems to be unused now and can get removed.

Done.

> The examples should be changed to use `esp-alloc` as a global allocator like here: https://github.com/esp-rs/esp-hal/blob/c7a236884589f33fe59b302154dfaa18c9af110c/examples/src/bin/wifi_dhcp.rs#L52

I'm not familiar with this. Should it just be added?
I don't see the examples explicitly using alloc, and from my testing it seems like `esp-mbedtls` use the `esp-wifi` memory management, which I think in version 0.9.1 is still tuned in cfg.toml .

I noticed things changed post 0.9.1 but `esp-mbedtls` is tied to specific `esp-hal` and `asp-wifi` versions, so things coming in later versions won't get updated automatically (unfortunately). 

Or maybe I'm missing something?

bjoernQ:
> I don't see the examples explicitly using alloc,

ah sorry - you are right that change is not yet released

bjoernQ:
I guess the fact the examples don't work currently because of the heap-size is a bit unfortunate - after the next release we have more RAM to use so maybe it's fine for now

@AnthonyGrondin what do you think?

yanshay:
The examples do work on `esp32s3` if I increase the heap size, but in the build it compiles the `esp32` and that compile fail with a larger heap size.
It is no different I think than before my PR since inside the `cfg.toml` it says to use  minimum but the file is below that minimum.
I can also say that with esp-wifi 0.9.1 I had to increase esp-wifi memory to get esp-mbedtls to work. Luckily for me I can use psram for my app heap so I have memory to spare. But I think a lower memory TLS is becoming really critical. I know its in work, just saying it could make esp-hal unusable for a lot of applications as TLS is not something one can do without if it is required.

bjoernQ:
> I can also say that with esp-wifi 0.9.1 I had to increase esp-wifi memory to get esp-mbedtls to work.

Yes since 0.9.0 a few things which used to be allocated statically are now allocated on the heap - so that's fine

AnthonyGrondin:
> I guess the fact the examples don't work currently because of the heap-size is a bit unfortunate - after the next release we have more RAM to use so maybe it's fine for now
> 
> @AnthonyGrondin what do you think?

I'm fine with any outcome. Some examples won't compile with `0.9.0` due to it being slightly larger, and with a static heap, we're at the maximum reach of what we can do with the default partition table.

I have somewhere a branch where I updated to the latest HEAD for `esp-hal` + `esp-wifi` to use the alloc system and the new init system, and it frees up a lot of memory, by sharing the same heap and not esp-wifi having its own static heap.

Once https://github.com/esp-rs/esp-hal/pull/2137 is in a new release, I'd like to tackle the issues of memory usage, now that it would be easier to do so.

Issue #42, opened 2024-08-12
Examples fail due to lack of memory


yanshay:
I tried to run the sync_client and async_client examples on ESP32, both had memory issues as below.

I tried to increase the esp_wifi heap size incrementally and tested, it always failed on memory allocation, until it couldn't compile anymore due to too large heap, reporting: section `.dram2_uninit' will not fit in region `dram2_seg'

Any way to get it to work? And even if I do, does it mean no memory will be left for my app? I don't need much but would need something.

Here are the execution logs:
```
INFO - esp-wifi configuration Config { rx_queue_size: 5, tx_queue_size: 3, static_rx_buf_num: 10, dynamic_rx_buf_num: 32, static_tx_buf_num: 0, dynamic_tx_buf_num: 32, ampdu_rx_enable: 0, ampdu_tx_enable: 0, amsdu_tx_enable: 0, rx_ba_win: 6, max_burst_size: 1, country_code: "CN", country_code_operating_class: 0, mtu: 1492, heap_size: 95000, tick_rate_hz: 100, listen_interval: 3, beacon_timeout: 6, ap_beacon_timeout: 300, failure_retry_cnt: 1, scan_method: 0 }
Call wifi_connect
Wait to get connected
Wait to get an ip address
Got ip Ok(IpInfo { ip: 192.168.10.96, subnet: Subnet { gateway: 192.168.10.1, mask: Mask(24) }, dns: Some(192.168.10.1), secondary_dns: None })
We are connected!
Making HTTP request
Start tls connect
WARN - Unable to allocate 1044 bytes
WARN - Unable to allocate 1044 bytes



!! A panic occured in 'examples/sync_client.rs', at line 125, column 33:
PanicInfo {
    payload: Any { .. },
    message: Some(
        called `Result::unwrap()` on an `Err` value: MbedTlsError(-9984),
    ),
    location: Location {
        file: "examples/sync_client.rs",
        line: 125,
        col: 33,
    },
    can_unwind: true,
    force_no_backtrace: false,
}

Backtrace:

0x400ee68b
0x400ee68b - core::result::unwrap_failed
    at /Users/my_user/.rustup/toolchains/esp/lib/rustlib/src/rust/library/core/src/result.rs:1679
0x400d69e5
0x400d69e5 - core::result::Result<T,E>::unwrap
    at /Users/my_user/.rustup/toolchains/esp/lib/rustlib/src/rust/library/core/src/result.rs:1102
0x400ec44c
[...]

PR #52, opened 2024-10-10
Update esp-hal to `0.21.0`. Update esp-wifi to `0.10.1`

AnthonyGrondin:
This updates all the `esp-*` dependencies to their latest version and use the new alloc system for Wifi allocations instead of a stack based allocator.

All examples compile fine, but we still need to test the runtime. I did a few tests locally.

@yanshay Can you check if this fixes the issues in https://github.com/esp-rs/esp-mbedtls/issues/42 ?

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -38,7 +38,7 @@ rustflags = [
 ]
 
 [unstable]
-build-std = ["core"]
+build-std = ["core", "alloc"]
 
 [alias]
 xtask = "run --manifest-path ./xtask/Cargo.toml --"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@v1
         with:
           target: riscv32imc-unknown-none-elf
-          toolchain: nightly-2024-06-12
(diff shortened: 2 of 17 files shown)


bjoernQ (approved):
Thanks!

The mTLS client example fails to connect currently for me even after updating the certificates but I don't expect that to be related to this PR

bjoernQ:
Lets wait for the confirmation regarding issue 42 - then I will merge this

yanshay:
I tried running both the sync_client and async_client examples.
They both compile but I can't get an IP no matter how many times I reset the device and how long I wait.
So I can't tell if it's working beyond that.
This WIFI issue is real severe for me and happens on several different devices (esp32, esp32s3).
But's that's to be continued on esp-hal discussions.

yanshay:
I tried running the async_client example on ESP32S3 and it's working.
The IP issues I have on ESP32. are reproduced also unrelated to esp-mbedtls.
So I don't know to say that #42 is really fixed, because that I tested on Esp32 and not on Esp32S3.
So seems like from my perspective (since I'm using mostly ESP32S3) this PR is fine.

bjoernQ:
I guess the DHCP are unrelated so going to merge this - thanks!

PR #75, opened 2025-05-14
wip: Share parsed `mbedtls_x509_crt` and `mbedtls_pk_context` between sessions

AnthonyGrondin:
Status: Still work in progress

The main intent behind this PR is to save memory allocated per session by re­using resources that could be shared between sessions, such as certificates.

This adds two new wrapper structs to initiate and parse X509 certs and Private keys, and manage the cleanup when they are dropped. Session now takes `Certificates` by reference so the same set of `Certificates` can be shared between sessions.

Also `Certificates` have been refactored to parse during init, and to contain the pointer to the allocated memory. One of my long term goal would be to be able to fetch data from the parsed certificates such as the expiring date, without needing extra external dependencies. The config struct is another few 200 bytes that could be saved by sharing it per session, but this would need to keep in memory what parameters were used. Or extend the `Certificates` builder mecanism so that `Certificates` own that config.

### DIFF
--- a/esp-mbedtls/src/edge_nal.rs
+++ b/esp-mbedtls/src/edge_nal.rs
@@ -10,7 +10,7 @@ use crate::{Certificates, Mode, TlsError, TlsReference, TlsVersion};
 pub struct TlsAcceptor<'d, T> {
     acceptor: T,
     min_version: TlsVersion,
-    certificates: Certificates<'d>,
+    certificates: &'d Certificates,
     tls_ref: TlsReference<'d>,
 }
 
@@ -29,7 +29,7 @@ where
     pub const fn new(
         acceptor: T,
         min_version: TlsVersion,
-        certificates: Certificates<'d>,
+        certificates: &'d Certificates,
         tls_ref: TlsReference<'d>,
     ) -> Self {
         Self {
@@ -78,7 +78,7 @@ pub struct TlsConnector<'d, T> {
     connector: T,
     servername: &'d CStr,
     min_version: TlsVersion,
-    certificates: Certificates<'d>,
+    certificates: &'d Certificates,
     tls_ref: TlsReference<'d>,
 }
 
@@ -99,7 +99,7 @@ where
         connector: T,
         servername: &'d CStr,
         min_version: TlsVersion,
-        certificates: Certificates<'d>,
+        certificates: &'d Certificates,
         tls_ref: TlsReference<'d>,
     ) -> Self {
         Self {
--- a/esp-mbedtls/src/lib.rs
+++ b/esp-mbedtls/src/lib.rs
@@ -244,12 +244,133 @@ impl<'a> X509<'a> {
     }
 }
 
-/// Certificates used for a connection.
-///
-/// # Note:
-/// Both [certificate](Certificates::certificate) and [private_key](Certificates::private_key) must be set in pair.
-#[derive(Debug, Clone, Copy, PartialEq)]
-pub struct Certificates<'a> {
+/// Creates a wrapper over [mbedtls_x509_crt] to safely manage allocation and freeing on drop
+#[derive(Debug)]
+struct MbedTLSX509Crt(*mut mbedtls_x509_crt);
+
+impl MbedTLSX509Crt {
+
+    fn new(certificate: X509<'_>) -> Result<Self, TlsError> {
+        unsafe {
+            let ptr = aligned_calloc(
+                align_of::<mbedtls_x509_crt>(),
+                size_of::<mbedtls_x509_crt>(),
+            ) as *mut mbedtls_x509_crt;
+            if ptr.is_null() {
+                return Err(TlsError::OutOfMemory);
+            }
+            mbedtls_x509_crt_init(ptr);
+
+            let cleanup = || {
+                mbedtls_x509_crt_free(ptr);
+                free(ptr as *const _);
+            };
+
+            match certificate.format {
+                CertificateFormat::PEM => {
+                    error_checked!(
+                        mbedtls_x509_crt_parse(ptr, certificate.as_ptr(), certificate.len()),
+                        cleanup
+                    )?;
+                }
+                CertificateFormat::DER => {
+                    error_checked!(
+                        // TODO: This currently requires the X509 to live for the duration of this
+                        // lifetime
+                        mbedtls_x509_crt_parse_der_nocopy(
+                            ptr,
+                            certificate.as_ptr(),
+                            certificate.len()
+                        ),
+                        cleanup
+                    )?;
+                }
+            }
+            Ok(Self(ptr))
+        }
+    }
+}
+
+impl core::ops::Deref for MbedTLSX509Crt {
+    type Target = *mut mbedtls_x509_crt;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Drop for MbedTLSX509Crt {
+    fn drop(&mut self) {
+        log::debug!("Dropping MbedTLSX509Crt");
+        unsafe {
+            mbedtls_x509_crt_free(self.0);
+            free(self.0 as *const _);
+        }
+    }
+}
+
+/// Creates a wrapper over [mbedtls_pk_context] to safely manage allocation and freeing on drop
+#[derive(Debug)]
+struct PkContext(*mut mbedtls_pk_context);
+
+impl PkContext {
+    fn new<'a>(private_key: X509<'a>, password: Option<&'a str>) -> Result<Self, TlsError> {
+        unsafe {
+            let ptr = aligned_calloc(
+                align_of::<mbedtls_pk_context>(),
+                size_of::<mbedtls_pk_context>(),
+            ) as *mut mbedtls_pk_context;
+            if ptr.is_null() {
+                return Err(TlsError::OutOfMemory);
+            }
+
+            mbedtls_pk_init(ptr);
+
+            let (password_ptr, password_len) = if let Some(password) = password {
+                (password.as_ptr(), password.len())
+            } else {
+                (core::ptr::null(), 0)
+            };
+            error_checked!(
+                mbedtls_pk_parse_key(
+                    ptr,
+                    private_key.as_ptr(),
+                    private_key.len(),
+                    password_ptr,
+                    password_len,
+                    None,
+                    core::ptr::null_mut(),
+                ),
+                || {
+                    mbedtls_pk_free(ptr);
+                    free(ptr as *const _);
+                }
+            )?;
+            Ok(Self(ptr))
+        }
+    }
+}
+
+impl core::ops::Deref for PkContext {
+    type Target = *mut mbedtls_pk_context;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Drop for PkContext {
+    fn drop(&mut self) {
+        log::warn!("Dropping PkContext");
+        unsafe {
+            mbedtls_pk_free(self.0);
+            free(self.0 as *const _);
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Certificates {
     /// Trusted CA (Certificate Authority) chain to be used for certificate
     /// verification during the SSL/TLS handshake.
     ///
@@ -265,7 +386,7 @@ pub struct Certificates<'a> {
     /// that will be used to verify the client's certificate during the handshake.
     /// When set to [None] the server will not request nor perform any verification
     /// on the client certificates. Only set when you want to use client authentication.
-    pub ca_chain: Option<X509<'a>>,
+    ca_chain: Option<MbedTLSX509Crt>,
 
     /// Own certificate chain used for requests
     /// It should contain in order from the bottom up your certificate chain.
@@ -279,35 +400,58 @@ pub struct Certificates<'a> {
     /// # Server:
     /// In server mode, this will be the certificate given to the client when
     /// performing a handshake.
-    pub certificate: Option<X509<'a>>,
+    certificate: Option<MbedTLSX509Crt>,
 
     /// Private key paired with the certificate. Must be set when [Certificates::certificate]
     /// is not [None]
-    pub private_key: Option<X509<'a>>,
-
-    /// Password used for the private key.
-    /// Use [None] when the private key doesn't have a password.
-    pub password: Option<&'a str>,
+    private_key: Option<PkContext>,
 }
 
-impl Default for Certificates<'_> {
+impl Default for Certificates {
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl Certificates<'_> {
+impl Certificates {
     /// Create a new instance of [Certificates] with no certificates whatsoever
     pub const fn new() -> Self {
         Self {
             ca_chain: None,
             certificate: None,
             private_key: None,
-            password: None,
         }
     }
 
-    // Initialize the SSL using this set of certificates
+    /// Initialize the certificate the own certificate chain used for requests
+    ///
+    /// # Errors
+    ///
+    /// This function will fail with [TlsError::OutOfMemory] if there's not enough memory to
+    /// allocate certificates.
+    pub fn with_certificates(
+        mut self,
+        certificate: X509<'_>,
+        private_key: X509<'_>,
+        password: Option<&'_ str>,
+    ) -> Result<Self, TlsError> {
+        self.certificate = Some(MbedTLSX509Crt::new(certificate)?);
+        self.private_key = Some(PkContext::new(private_key, password)?);
+        Ok(self)
+    }
+
+    /// Initialize the Certificate Authority chain used for requests
+    ///
+    /// # Errors
+    ///
+    /// This function will fail with [TlsError::OutOfMemory] if there's not enough memory to
+    /// allocate certificates.
+    pub fn with_ca_chain(mut self, ca_chain: X509<'_>) -> Result<Self, TlsError> {
+        self.ca_chain = Some(MbedTLSX509Crt::new(ca_chain)?);
+        Ok(self)
+    }
+
+    /// Initialize the SSL using this set of certificates
     fn init_ssl(
         &self,
         mode: Mode,
@@ -317,19 +461,9 @@ impl Certificates<'_> {
             *mut mbedtls_ctr_drbg_context,
             *mut mbedtls_ssl_context,
             *mut mbedtls_ssl_config,
-            *mut mbedtls_x509_crt,
-            *mut mbedtls_x509_crt,
-            *mut mbedtls_pk_context,
         ),
         TlsError,
     > {
-        // Make sure that both certificate and private_key are either Some() or None
-        assert_eq!(
-            self.certificate.is_some(),
-            self.private_key.is_some(),
-            "Both certificate and private_key must be Some() or None"
-        );
-
         // TODO: Allocate a lot of these things:
         // - In one chunk
         // - With a `Box`, which is safer
@@ -370,50 +504,8 @@ impl Certificates<'_> {
                 return Err(TlsError::OutOfMemory);
             }
 
-            let crt = aligned_calloc(
-                align_of::<mbedtls_x509_crt>(),
-                size_of::<mbedtls_x509_crt>(),
-            ) as *mut mbedtls_x509_crt;
-            if crt.is_null() {
-                free(drbg_context as *const _);
-                free(ssl_context as *const _);
-                free(ssl_config as *const _);
-                return Err(TlsError::OutOfMemory);
-            }
-
-            let certificate = aligned_calloc(
-                align_of::<mbedtls_x509_crt>(),
-                size_of::<mbedtls_x509_crt>(),
-            ) as *mut mbedtls_x509_crt;
-            if certificate.is_null() {
-                free(drbg_context as *const _);
-                free(ssl_context as *const _);
-                free(ssl_config as *const _);
-                free(crt as *const _);
-                return Err(TlsError::OutOfMemory);
-            }
-
-            let private_key = aligned_calloc(
-                align_of::<mbedtls_pk_context>(),
-                size_of::<mbedtls_pk_context>(),
-            ) as *mut mbedtls_pk_context;
-            if private_key.is_null() {
-                free(drbg_context as *const _);
-                free(ssl_context as *const _);
-                free(ssl_config as *const _);
-                free(crt as *const _);
-                free(certificate as *const _);
-                return Err(TlsError::OutOfMemory);
-            }
-
             mbedtls_ssl_init(ssl_context);
             mbedtls_ssl_config_init(ssl_config);
-            // Initialize CA chain
-            mbedtls_x509_crt_init(crt);
-            // Initialize certificate
-            mbedtls_x509_crt_init(certificate);
-            // Initialize private key
-            mbedtls_pk_init(private_key);
 
             //(*ssl_config).private_f_dbg = Some(dbg_print);
             mbedtls_ssl_conf_dbg(ssl_config, Some(dbg_print), core::ptr::null_mut());
@@ -428,15 +520,9 @@ impl Certificates<'_> {
                 mbedtls_ctr_drbg_free(drbg_context);
                 mbedtls_ssl_config_free(ssl_config);
                 mbedtls_ssl_free(ssl_context);
-                mbedtls_x509_crt_free(crt);
-                mbedtls_x509_crt_free(certificate);
-                mbedtls_pk_free(private_key);
                 free(drbg_context as *const _);
                 free(ssl_context as *const _);
                 free(ssl_config as *const _);
-                free(crt as *const _);
-                free(certificate as *const _);
-                free(private_key as *const _);
             };
 
             error_checked!(
@@ -471,66 +557,16 @@ impl Certificates<'_> {
                 )?;
             }
 
-            if let Some(ca_chain) = self.ca_chain {
-                error_checked!(
-                    mbedtls_x509_crt_parse(crt, ca_chain.as_ptr(), ca_chain.len()),
-                    cleanup
-                )?;
+            if let (Some(certificate), Some(private_key)) = (&self.certificate, &self.private_key) {
+                mbedtls_ssl_conf_own_cert(ssl_config, **certificate, **private_key);
             }
 
-            if let (Some(cert), Some(key)) = (self.certificate, self.private_key) {
-                // Certificate
-                match cert.format {
-                    CertificateFormat::PEM => {
-                        error_checked!(
-                            mbedtls_x509_crt_parse(certificate, cert.as_ptr(), cert.len()),
-                            cleanup
-                        )?;
-                    }
-                    CertificateFormat::DER => {
-                        error_checked!(
-                            mbedtls_x509_crt_parse_der_nocopy(
-                                certificate,
-                                cert.as_ptr(),
-                                cert.len(),
-                            ),
-                            cleanup
-                        )?;
-                    }
-                }
-
-                // Private key
-                let (password_ptr, password_len) = if let Some(password) = self.password {
-                    (password.as_ptr(), password.len())
-                } else {
-                    (core::ptr::null(), 0)
-                };
-                error_checked!(
-                    mbedtls_pk_parse_key(
-                        private_key,
-                        key.as_ptr(),
-                        key.len(),
-                        password_ptr,
-                        password_len,
-                        None,
-                        core::ptr::null_mut(),
-                    ),
-                    cleanup
-                )?;
-
-                mbedtls_ssl_conf_own_cert(ssl_config, certificate, private_key);
+            if let Some(ref ca_chain) = self.ca_chain {
+                mbedtls_ssl_conf_ca_chain(ssl_config, **ca_chain, core::ptr::null_mut());
             }
 
-            mbedtls_ssl_conf_ca_chain(ssl_config, crt, core::ptr::null_mut());
             error_checked!(mbedtls_ssl_setup(ssl_context, ssl_config), cleanup)?;
-            Ok((
-                drbg_context,
-                ssl_context,
-                ssl_config,
-                crt,
-                certificate,
-                private_key,
-            ))
+            Ok((drbg_context, ssl_context, ssl_config))
         }
     }
 }
@@ -671,9 +707,6 @@ pub struct Session<'a, T> {
     drbg_context: *mut mbedtls_ctr_drbg_context,
     ssl_context: *mut mbedtls_ssl_context,
     ssl_config: *mut mbedtls_ssl_config,
-    crt: *mut mbedtls_x509_crt,
-    client_crt: *mut mbedtls_x509_crt,
-    private_key: *mut mbedtls_pk_context,
     state: SessionState,
     _tls_ref: TlsReference<'a>,
 }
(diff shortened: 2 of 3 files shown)


esp-mbedtls/src/edge_nal.rs
```diff
@@ -10,7 +10,7 @@ use crate::{Certificates, Mode, TlsError, TlsReference, TlsVersion};
 pub struct TlsAcceptor<'d, T> {
     acceptor: T,
     min_version: TlsVersion,
-    certificates: Certificates<'d>,
```
ivmarkov, line 13: I think the signature here and everywhere below should rather be `certificates: &'d Certificates<'d>`. This way:
- If the `Certificates` type needs to preserve a reference to `X509`, it can do it
- If it does not have to, it will essentially be `Certificates<'static>` which - due to the covariance of `'d` would be represented just fine with `&'d Certificates<'d>`

esp-mbedtls/src/lib.rs
```diff
@@ -244,12 +244,133 @@ impl<'a> X509<'a> {
...
 
-/// Certificates used for a connection.
-///
-/// # Note:
-/// Both [certificate](Certificates::certificate) and [private_key](Certificates::private_key) must be set in pair.
-#[derive(Debug, Clone, Copy, PartialEq)]
-pub struct Certificates<'a> {
+/// Creates a wrapper over [mbedtls_x509_crt] to safely manage allocation and freeing on drop
+#[derive(Debug)]
+struct MbedTLSX509Crt(*mut mbedtls_x509_crt);
```
ivmarkov, line 249: Changing this to 
```rust
MbedTlsX509Crt<'d> {
    crt: *mut mbedtls_x509_crt,
    _t: PhantomData<&'d ()>,
}
```

... would help express the fact that the `*mut mbedtls_x509_crt` _might_ be keeping a reference to a `X509` cert contents.

}

esp-mbedtls/src/lib.rs
```diff
@@ -244,12 +244,133 @@ impl<'a> X509<'a> {
...
+
+impl MbedTLSX509Crt {
+
+    fn new(certificate: X509<'_>) -> Result<Self, TlsError> {
```
ivmarkov, line 253: To fix the "TODO" inside this methods, we actually need two methods:

A copying variant, for PEM and perhaps for DER too where the user would not like to keep a ref to the `x509` contents:
```rust
impl MbedTlsX509Crt<'static> {
    fn new(x509: X509<'_>) -> Result<Self, TlsError> { ... }
    ...
```

A non-copying variant (would only work for DER X509:
```rust
impl<'d> MbedTlsX509Crt<'d> {
    fn new(x509: X509<'d>) -> Result<Self, TlsError> { ... }
    ...
}
```

esp-mbedtls/src/lib.rs
```diff
@@ -244,12 +244,133 @@ impl<'a> X509<'a> {
...
+                        cleanup
+                    )?;
+                }
+            }
+            Ok(Self(ptr))
+        }
+    }
+}
+
+impl core::ops::Deref for MbedTLSX509Crt {
```
ivmarkov, line 294: Not sure we need this actually. Implementing `Deref` is always controversial (though it gives extra usability, but then - for an internal struct?...)

esp-mbedtls/src/lib.rs
```diff
@@ -244,12 +244,133 @@ impl<'a> X509<'a> {
...
+        log::warn!("Dropping PkContext");
+        unsafe {
+            mbedtls_pk_free(self.0);
+            free(self.0 as *const _);
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Certificates {
```
ivmarkov, line 373: `Certificates<'d>`

esp-mbedtls/src/lib.rs
```diff
@@ -244,12 +244,133 @@ impl<'a> X509<'a> {
...
+                    mbedtls_pk_free(ptr);
+                    free(ptr as *const _);
+                }
+            )?;
+            Ok(Self(ptr))
+        }
+    }
+}
+
+impl core::ops::Deref for PkContext {
```
ivmarkov, line 354: Ditto.

esp-mbedtls/src/lib.rs
```diff
@@ -265,7 +386,7 @@ pub struct Certificates<'a> {
     /// that will be used to verify the client's certificate during the handshake.
     /// When set to [None] the server will not request nor perform any verification
     /// on the client certificates. Only set when you want to use client authentication.
-    pub ca_chain: Option<X509<'a>>,
+    ca_chain: Option<MbedTLSX509Crt>,
```
ivmarkov, line 389: `MbedTlsX509Crt<'d>`

esp-mbedtls/src/lib.rs
```diff
@@ -279,35 +400,58 @@ pub struct Certificates<'a> {
     /// # Server:
     /// In server mode, this will be the certificate given to the client when
     /// performing a handshake.
-    pub certificate: Option<X509<'a>>,
+    certificate: Option<MbedTLSX509Crt>,
```
ivmarkov, line 403: Ditto - `MbedTlsX509Crt<'d>`

esp-mbedtls/src/lib.rs
```diff
@@ -279,35 +400,58 @@ pub struct Certificates<'a> {
...
+impl Default for Certificates {
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl Certificates<'_> {
+impl Certificates {
     /// Create a new instance of [Certificates] with no certificates whatsoever
     pub const fn new() -> Self {
```
ivmarkov, line 418: Is this a real use-case? Can the library actually operate without any certs?

esp-mbedtls/src/lib.rs
```diff
@@ -279,35 +400,58 @@ pub struct Certificates<'a> {
...
     }
 
-    // Initialize the SSL using this set of certificates
+    /// Initialize the certificate the own certificate chain used for requests
+    ///
+    /// # Errors
+    ///
+    /// This function will fail with [TlsError::OutOfMemory] if there's not enough memory to
+    /// allocate certificates.
+    pub fn with_certificates(
```
ivmarkov, line 432: This would need two versions, just like `MbedTlsX509Crt` has two versions - one, in `Cetificates<'static>` which does copy, and another - in `Certificates<'d>` which tries to go non-copy where it can.

esp-mbedtls/src/lib.rs
```diff
@@ -279,35 +400,58 @@ pub struct Certificates<'a> {
...
+        Ok(self)
+    }
+
+    /// Initialize the Certificate Authority chain used for requests
+    ///
+    /// # Errors
+    ///
+    /// This function will fail with [TlsError::OutOfMemory] if there's not enough memory to
+    /// allocate certificates.
+    pub fn with_ca_chain(mut self, ca_chain: X509<'_>) -> Result<Self, TlsError> {
```
ivmarkov, line 449: Ditto.

esp-mbedtls/src/lib.rs
```diff
@@ -265,7 +386,7 @@ pub struct Certificates<'a> {
     /// that will be used to verify the client's certificate during the handshake.
     /// When set to [None] the server will not request nor perform any verification
     /// on the client certificates. Only set when you want to use client authentication.
```
ivmarkov, line 388: "only set this when you want to use client authentication".
Hmm, is this correct? My understanding is that if you are running a TLS client (say - an HTTPS client), you might still need a bunch of CA certificates so as to verify the certificate that the server is providing to you? I don't think the server is obliged at all to give you the CA cert, it only gives you its own cert signed by the CA cert but then the CA cert (in fact - a whole bundle of these) needs to be configured in the client?

Where I'm going with that is that both client and server (the server needs it when client auth is enabled) need (optional) access to something that ESP-IDF calls a "certificate bundle" which is - to oversimplify - a `X509[]` array which is just a bunch of (CA or interim) certificates.

I think we are kind of missing the whole notion of a "certificate bundle" actually... Also, the certificate bundle is always the same and shared across everything, so perhaps it needs to be modeled with its own type. `CertificatesBundle` or something.

esp-mbedtls/src/lib.rs
ivmarkov: The lifetimes are covariant though so **a single lifetime would do**. If you want to keep the builder pattern, `new` would return a `Certificates<'static>` which would either stay like that (copy) or would be changed to `Certificates<'a>` by a builder non-copy method, where `'a` would be a generic lifetime capturing the actual x509 lifetime (which can _still_ end up being static after all).

esp-mbedtls/src/lib.rs
AnthonyGrondin: This seems to be a documentation error. It is indeed required nevertheless, and the optional certificate part for a server should actually be for server certficates, and not CA chain.

Currently X509 seems to support passing multiple chained certificates in the same parsing to create a bundle, but we might also extend the API to allow passing in a list of `X509[]` if the need is there.

ivmarkov: > Currently X509 seems to support passing multiple chained certificates in the same parsing to create a bundle, but we might also extend the API to allow passing in a list of X509[] if the need is there.

My understanding is that a certificate chain (each cert in the chain signed by the following one) is a different thing from a bundle, which is a set of disparate root certs, as in google's, Mozilla's, verisign's etc. but I could be wrong.

AnthonyGrondin: You could be right. I don't have much knowledge on the bundle properties, and even less knowledge about how `ESP-IDF` implements it.

AnthonyGrondin:
I added `no-copy` variants for parsing. I wish Rust could let us re-use the same function signature with different lifetimes requirements but its architecture doesn't allow it.

I also exposed the certificate verification mode to Session so that users can set it. It is optional because sane defaults exists for this setting. That being said, that makes the `Session` constructor pretty bloated, and I wonder if we should maybe switch to a builder pattern in a near-distant future. That would also avoid making a breaking API change every single iteration we try to change so feature around.

Currently client examples fail with verification failed so I have to figure out why that is. Server examples work fine.

AnthonyGrondin:
So I'm still unsatisfied with the general API of this PR and general project. I refactored a proprietary app we're using, and I still have the following concerns:

 - `Session::new()` now has an ugly optional parameter in the middle of its argument, and it's a breaking change.
   We could either:
   1. Move the setting to `Certificates`, but this will block a user from using the same set of certificates with two different verification modes, and this is a session setting, not certificate setting.
   2. Provide a `set_authmode()` API to session to programatically change the authmode of `mbedtls_ssl_config` BUT `Session:new` already call `init_ssl()` which calls `mbedtls_ssl_setup()` with the config, and the documentation specifies
   > No copy of the configuration context is made, it can be shared by many mbedtls_ssl_context structures.
   The conf structure will be accessed during the session. It must not be modified or freed as long as the session is active.
   
   So we cannot update the config once it has been used in a setup.
   
   3. Refactor `Session` to use a builder pattern, but this creates extra code and memory usage, and the goal is already to trim down both, so I consider this a regression in a way.
- `MbedTLSX509Crt` and `PkContext` are still unexposed, and so, the user still doesn't have access to the parsed certificate and its parsed data, which was the original intent. My use case is that I want to be able to see the expiring date of a certificate without needing an extra dependency. It's already all available in there.

AnthonyGrondin:
> Why is it optional? What does passing `None` to auth_mode even mean, especially given that `AuthMode::None` also exists?

I agree that the current API is counter intuitive. Auth_mode should be optional because there are sane defaults for this setting which are `MBEDTLS_SSL_VERIFY_NONE` for server and `MBEDTLS_SSL_VERIFY_REQUIRED` for client. And these should be used by default for 90% of the time for security and usability purpose.

There's `MBEDTLS_SSL_VERIFY_OPTIONAL` that exists but since we don't expose `mbedtls_ssl_get_verify_result()` it's pretty much useless and the equivalent of `MBEDTLS_SSL_VERIFY_NONE` as-is, but it's still there.

A user would only really want to change these for:
1. (Client) => use `MBEDTLS_SSL_VERIFY_NONE` to disable verification, but that doesn't work for TLS1.3 per my experimentations
2. (Server) => use `MBEDTLS_SSL_VERIFY_REQUIRED` to request and force the peer to send client certificates to allow the connection.

> Isn't this just a matter of putting in some extra code in Certificate so that you have access to the data you need? Is this related in any way to the other changes?

Yeah, I think we can use getters on Certificates to return the pointers directly, wrapped in options `Option<*mut mbedtls_x509_crt>` or `Option<&MbedTLSX509Crt<'_>>` for `ca_chain` and `certificate`

ivmarkov:
> > Why is it optional? What does passing `None` to auth_mode even mean, especially given that `AuthMode::None` also exists?
> 
> I agree that the current API is counter intuitive. Auth_mode should be optional because there are sane defaults for this setting which are `MBEDTLS_SSL_VERIFY_NONE` for server and `MBEDTLS_SSL_VERIFY_REQUIRED` for client. And these should be used by default for 90% of the time for security and usability purpose.
> 
> There's `MBEDTLS_SSL_VERIFY_OPTIONAL` that exists but since we don't expose `mbedtls_ssl_get_verify_result()` it's pretty much useless and the equivalent of `MBEDTLS_SSL_VERIFY_NONE` as-is, but it's still there.
> 
> A user would only really want to change these for:
> 
> 1. (Client) => use `MBEDTLS_SSL_VERIFY_NONE` to disable verification, but that doesn't work for TLS1.3 per my experimentations
> 2. (Server) => use `MBEDTLS_SSL_VERIFY_REQUIRED` to request and force the peer to send client certificates to allow the connection.

If we want a cleaner api, we might try fusing `Mode` and `AuthMode` into a single thing then, as to me it seems these are co-related and might be cleaner to express together?

AnthonyGrondin:
Good idea. Perhaps a `SessionConfig` struct which would contain `Mode`, `AuthMode` and `TlsVersion`. If we ever add extra parameters, we can extend it

AnthonyGrondin:
I'm getting pretty happy with the result so far.

- Better lifetime management of X509 and Private key
- Introduction of `SessionConfig` and user configurable `AuthMode` for better usability
- Getters to fetch the parsed X509 properties
- Certificate builder API that allows copy and no-copy variants.

One edge-case I found is that `Certificates::with_certificates(&self, cert, private_key)` which makes it easy to fulfill the requirement that both certificate and private key be used at the same time, or not, takes `X509<'_>` for both of its fields, requiring both the certificate and private key to be present in RAM at the same time.

I've calculated the memory cost difference of this to be about the size of the certificate slice, 2048KB in the example below.

That is:
```
-> Certificate loaded from RAM: 2048KB
-> Private key loaded from RAM: 2048KB
--- Total RAM usage: 4096KB
-> Copy parse: Certificates::with_certificates(&self, cert, private_key)
-> Allocation of `mbedtls_x509_crt` + parse: Figuratively 2048KB alloc
-> Allocation of `mbedtls_pk_context` + parse: Let's say 3000KB alloc
--- Total RAM usage: 9144KB
-> Drop of unused certificate and private key slice
---- Total RAM usage: 5048KB 
```

While if they were parsed one after the other, that would be something like:
```
-> Certificate loaded from RAM: 2048KB
-> Copy parse X509 certificate only
-> Allocation of `mbedtls_x509_crt` + parse: Figuratively 2048KB alloc
--- Total RAM usage: 4096KB
-> Drop unused certificate slice
--- Total RAM usage: 2048KB
-> Private key loaded from RAM: 2048KB
-> Copy parse of X509 private key
-> Allocation of `mbedtls_pk_context` + parse: Let's say 3000KB
--- Total RAM usage: 7096KB
-> Drop unused private key slice
---- Total RAM usage: 5048KB 
```

If a user really wants to save those bytes, they can use the nocopy variant.

esp-mbedtls/src/lib.rs
```diff
@@ -698,21 +958,16 @@ impl<'a, T> Session<'a, T> {
     /// invalid format.
     pub fn new(
         stream: T,
-        mode: Mode,
-        min_version: TlsVersion,
-        certificates: Certificates,
+        config: SessionConfig<'_>,
+        certificates: &Certificates,
```
ivmarkov, line 962: This **must** be `&'a Certificates<'a>` or else we are set for big problems :)

esp-mbedtls/src/lib.rs
```diff
@@ -1014,21 +1261,16 @@ pub mod asynch {
         /// invalid format.
         pub fn new(
             stream: T,
-            mode: Mode,
-            min_version: TlsVersion,
-            certificates: Certificates,
+            config: SessionConfig<'_>,
+            certificates: &Certificates,
```
ivmarkov, line 1265: Ditto: `&'a Certificates<'a>`

ivmarkov (changes requested):

esp-mbedtls/src/lib.rs
```diff
@@ -244,12 +320,195 @@ impl<'a> X509<'a> {
...
+                size_of::<mbedtls_x509_crt>(),
+            ) as *mut mbedtls_x509_crt;
+            if ptr.is_null() {
+                return Err(TlsError::OutOfMemory);
+            }
+            mbedtls_x509_crt_init(ptr);
+
+            let cleanup = || {
+                mbedtls_x509_crt_free(ptr);
+                free(ptr as *const _);
```
ivmarkov, line 354: `mbedtls_free`

esp-mbedtls/src/lib.rs
```diff
@@ -244,12 +320,195 @@ impl<'a> X509<'a> {
...
+    ///
+    /// This will return an error if an error occurs during parsing.
+    /// [TlsError::InvalidFormat] will be returned if a PEM encoded certificate is passed.
+    fn new_no_copy(certificate: X509<'d>) -> Result<Self, TlsError> {
+        // Currently no copy is only supported by DER certificates
+        if matches!(certificate.format(), CertificateFormat::PEM) {
+            return Err(TlsError::InvalidFormat);
+        }
+        unsafe {
+            let ptr = aligned_calloc(
```
ivmarkov, line 405: `mbedtls_calloc`

esp-mbedtls/src/lib.rs
```diff
@@ -244,12 +320,195 @@ impl<'a> X509<'a> {
...
+            })
+        }
+    }
+}
+
+impl Drop for MbedTLSX509Crt<'_> {
+    fn drop(&mut self) {
+        unsafe {
+            mbedtls_x509_crt_free(self.crt);
+            free(self.crt as *const _);
```
ivmarkov, line 443: `mbedtls_free`

esp-mbedtls/src/lib.rs
```diff
@@ -244,12 +320,195 @@ impl<'a> X509<'a> {
...
+                    private_key.as_ptr(),
+                    private_key.len(),
+                    password_ptr,
+                    password_len,
+                    None,
+                    core::ptr::null_mut(),
+                ),
+                || {
+                    mbedtls_pk_free(ptr);
+                    free(ptr as *const _);
```
ivmarkov, line 493: `mbedtls_free`

esp-mbedtls/src/lib.rs
```diff
@@ -428,21 +755,15 @@ impl Certificates<'_> {
                 mbedtls_ctr_drbg_free(drbg_context);
                 mbedtls_ssl_config_free(ssl_config);
                 mbedtls_ssl_free(ssl_context);
-                mbedtls_x509_crt_free(crt);
-                mbedtls_x509_crt_free(certificate);
-                mbedtls_pk_free(private_key);
                 free(drbg_context as *const _);
```
ivmarkov, line 758: `mbedtls_free` instead of just `free` on all of these

esp-mbedtls/src/lib.rs
```diff
@@ -911,15 +1166,9 @@ impl<T> Drop for Session<'_, T> {
             mbedtls_ctr_drbg_free(self.drbg_context);
             mbedtls_ssl_config_free(self.ssl_config);
             mbedtls_ssl_free(self.ssl_context);
-            mbedtls_x509_crt_free(self.crt);
-            mbedtls_x509_crt_free(self.client_crt);
-            mbedtls_pk_free(self.private_key);
             free(self.drbg_context as *const _);
```
ivmarkov, line 1169: `mbedtls_free` instead of just `free` on all of these.

esp-mbedtls/src/lib.rs
```diff
@@ -1045,15 +1287,9 @@ pub mod asynch {
                 mbedtls_ctr_drbg_free(self.drbg_context);
                 mbedtls_ssl_config_free(self.ssl_config);
                 mbedtls_ssl_free(self.ssl_context);
-                mbedtls_x509_crt_free(self.crt);
-                mbedtls_x509_crt_free(self.client_crt);
-                mbedtls_pk_free(self.private_key);
                 free(self.drbg_context as *const _);
```
ivmarkov, line 1290: `mbedtls_free` instead of just `free` on all of these.

ivmarkov (changes requested):

ivmarkov:
I've added feedback regarding the lifetimes of `Certtificates` on `Session::new`.

Also, I don't see why we should delay replacing `aligned_alloc` and `free` with `mbedtls_calloc` and `mbedtls_free`. We should no longer be using the global heap (neither the Rust one, nor the C one) in any of `esp-mbedtls`. Or else - and at the very least - I cannot make use of `esp-mbedtls` inside [`openthread`](https://github.com/esp-rs/openthread).

AnthonyGrondin:
Thank you very much for your review.

I replaced `calloc` with `mbedtls_calloc` inside of `aligned_calloc`, but do we sill need it for memory alignment?

yanshay:
I hope you don't mind me jumping in, but a small question I think is related here. Will the changes you make will allow utilizing PSRAM memory for esp-mbedtls?

Previous version afaik used the esp-wifi allocations and therefore the DRAM while technically I understand it could use PSRAM for at least some of the operations (unlike esp-wifi that must use DRAM).

ivmarkov:
> Thank you very much for your review.
> 
> I replaced `calloc` with `mbedtls_calloc` inside of `aligned_calloc`, but do we sill need it for memory alignment?

Great question that does not have a simple answer.

The TL;DR is: it should work without explicit alignment. This is so because `malloc`/`calloc` are required to return memory which is aligned to the biggest native C type supported in hardware that needs aligned RW. In other words, if e.g. `u64` for your platform needs to be aligned at the 8-byte boundary (but e.g. `u128` isn't), `malloc` for that platform is **required** to always return memory aligned to 8 bytes, regardless of how much memory you request it to allocate.

This works great for RiscV and Cortex M where `malloc` returns 4-bytes-aligned memory, and there only `u32` needs to be aligned, bigger types don't have to.

For `xtensa`, it is a mess. While the `xtensa` spec **does** require aligned `u64` (not sure for `u128`) and the LLVM `xtensa` `rustc` would even panic-complain (since recently, even in `release`!) if you try to dereference an un-aligned `u64`, `esp-wifi`'s (or is it now `esp-malloc`s?) `malloc` emulation only alignes to 4 bytes, which is an omission I recently mentioned to @bjoernQ .

With that said, The C compiler (at least in release) does NOT complain on unaligned `u64`s and the CPU happily executes these unaligned `u64` RWs just fine. Also, the ESP-IDF `malloc` impl also implements only 4-byte alignment and they say if you want higher alignment, you should use `posix_memalign` instead (probably they did it to conserve memory), but this way they violate the `malloc` contract.


Anyway, long story short, I say let's just use `mbedtls_calloc` (and thus potentially 4-bytes `malloc`) everywhere on `xtensa` too, unconditionally. As long as the C mbedtls structs do not contain `u64`s OR you do not dereference them from Rust, we should be fine. Should we see a panic in some of our own MbedTLS structures' allocations (to the others we don't have control) we either have to ask `esp-wifi` to align to 8 bytes, or we should implement our own little alignment routine on top of `mbedtls_calloc`.

ivmarkov:
@AnthonyGrondin  BTW if this is ready from your POV shall I merge?

AnthonyGrondin:
Code seems ready but we still need to do some testing.

I think the client examples broke and I need to figure out why.

ivmarkov:
@AnthonyGrondin I think before thinking about connection splitting, we probably need to merge this one first...

AnthonyGrondin:
I rebased on top of master and tested the examples on **esp32s3**.

mTLS client examples are still broken but they were broken on `main` so I don't think this is a bug that this PR introduces.

AnthonyGrondin:
One last thing I'd like to see implemented here, is sharing certificates across different `Certificates<'a>` instances. Copying / Cloning the whole `Certificates` here is pointless because we pretty much pass it by reference everywhere, but there are cases where one would want to reuse the same **CA chain** for example, with another `Certificates<'a>` instance.

I have a case where I renew certificates on a device, so both certificates and private key are different, but the **CA Chain** is the same, and it would be a waste of memory to create and alloc another instance, if one exists already in memory.

I'm still trying to think how it could be implemented. Since we use alloc anyways, maybe some internal RC

AnthonyGrondin:
> > I'm still trying to think how it could be implemented. Since we use alloc anyways, maybe some internal RC
> 
> ... except we actually no longer use `alloc`. Why not passing the CA chain as a regular reference into `Certificates`? This way you can share it trivially across.
> 
> ( BTW: All of this sharing can only happen within a single thread. (Not that this is important on bare-metal.). Just mentioning. [The restriction is within the mbedtls C lib as well, I think, in that you can't share its structures across threads](https://github.com/fortanix/rust-mbedtls?tab=readme-ov-file#change-of-status). At least not anymore. )

I did a local refactor, removing the no_copy variants of `Certificates` and making `with_certificates() / with_ca_chain()` take `&'d MbedTLSX509Crt<'_>` (I did the same for PkContext) and store it within Certificates, so it basically becomes a facade. It works for simple examples such as the examples in this library, but for a more complex case such as within the app I'm developping, I get lifetime issues, because the underlying certificates `MbedTLSX509Crt<'_>` need to live for longer. It's harder to pass `Certificates<'_>` to different scopes.

Maybe I could derive `Copy` on `Certificates` and make `Session` take it owned so that the constructor looks something like the following, and `Certificates` is only a facade:

```Rust
let mut session = Session::new(
    &mut socket,
    SessionConfig::new(
        Mode::Client {
            servername: SERVERNAME,
        },
        TlsVersion::Tls1_3,
    ),
    Certificates::new()
        .with_certificates(&crt, &pk)
        .with_ca_chain(&ca_chain),
    tls.reference(),
)
```

and let the user deal with the lifetime management of 3 different variables, instead of a single one.

AnthonyGrondin:
I pushed the refactor to https://github.com/esp-rs/esp-mbedtls/tree/refactor/x509-memory-improvement-test

[diff](https://github.com/AnthonyGrondin/esp-mbedtls/compare/9e871301220937b8c374592ff2801e4d03b31aee...217f6d62a42b514ea0896362bbfdb6adf89c961a)

I changed the `edge_server` example to reproduce something similar to what I'm doing in my application. It would work when `Certificates` owned the underlying certs, but now I would need to "Box::leak()" all the variants.

ivmarkov:
> I pushed the refactor to https://github.com/esp-rs/esp-mbedtls/tree/refactor/x509-memory-improvement-test
> 
> [diff](https://github.com/AnthonyGrondin/esp-mbedtls/compare/9e871301220937b8c374592ff2801e4d03b31aee...217f6d62a42b514ea0896362bbfdb6adf89c961a)
> 
> I changed the `edge_server` example to reproduce something similar to what I'm doing in my application. It would work when `Certificates` owned the underlying certs, but now I would need to "Box::leak()" all the variants.

You mean this line: https://github.com/esp-rs/esp-mbedtls/blob/217f6d62a42b514ea0896362bbfdb6adf89c961a/examples/edge_server.rs#L162?

Yes, it can't work, but I wonder how `Box::new` (previous line) actually even worked for your case? If I'm not mistaken (could be) Box<T> requires `where T: 'static` and your `Certificates` are non-static as they hold references to the stack-allocated `crt` and `private_key`.

A few questions/comments:
- Why do you want to `Box` `Certificates`  in the first place? and why do you want `'static Certificates` too?
- If you want a `'static` ref to `Certificates` (for whatever the reason is), then the refs _inside_ `Certificates` must be static too. In other words, either you box-allocate `crt` and `private_key` and then leak them into static, or you just `mk_static`  them prior to pushing their refs into `Certificates`. In either case you want `Certificates<'static>` so that when you leak it (or `mk_static` it) it would work and you'll get `&'static Certificates<'static>`

ivmarkov:
... so you could even roll-your-own type-alias: `pub type RefCtCertificates<Rc<MbedTLSX509Crt>, Rc<PkContext>>.

But the point is, `esp-mbedtls` does not know (or care) about you using `alloc` still.

AnthonyGrondin:
My use case is a bit unsafe to starts with. I wrap the `HttpClient` from https://github.com/drogue-iot/reqwless and I have a function to renew the certificates, on this wrapper, which takes `&'_ mut HttpClient<'a, T, D>` where I swap the `TlsConfig` to use my new certs, then make a call with my API to validate that they work and then:
1. If failed, re-use the original certs, which stays unchanged
2. Save the certs in flash storage, and update a certain static where they are stored.

I haven't found a safe easy way to do it but so far this seems to work for me. I only need to improve the memory usage.

I think in my final implementation I'll use a critical section or some blocking async to ensure that the client isn't used elsewhere during this certificate swap trickery. I could also create another client instance, but my current architecture uses a singleton for simplicity.

ivmarkov:
@AnthonyGrondin sorry for the delay, and give me one more day. It's been a long time, and now I need to dedicate a few hours to get back into the topic.

Will try to do it tmr morning.

ivmarkov:
> My reasoning came that as much as I tried to make it work with `*mut mbedtls_x509_crt`, having an inner wrapper similar to `Cow`, 

I no longer remember anymore from where the whole `Cow` came to be? I personally never needed it and all the time was talking about abstracting over references with `Borrow` I think. Which is a _completely different thing_. In the name of simplicity and moving this forward, I suggest we forget about `Cow` and `Borrow`. For now.

> we ultimately hide potentially unsafe, / UB behaviour, by sharing a mutable pointer. I checked; Does it actually need to be mutable?

The pointers are `*mut` simply because this is what C expects and that's because MbedTLS does not know any better. It wants `*mut` because C allows this, and moreover it allows aliased mutable access, so they did not bother chaning these to `*const`.

But they treat them as `*const` or else we should forget about certificate and priv-key sharing amongst sessions.

> It seems initialisation context don't update the pointers (I could be very wrong), so we can safely pass `*const mbedtls_x509_crt` crafted from `&mbedtls_x509_crt`.
> 
> So overall we just keep pointers wrappers, which's sole purpose is to "safely-_ish_" initialize the pointers, and ensure they properly get deallocated on drop. Meanwhile, we focus on exposing the types directly: `mbedtls_x509_crt / mbedtls_pk_context` which is closer to my initial intents.

T.b.h. I'm not really sure what this latest change is giving us as a benefit. Basically, it is more or less like the previous state _except_ that now you have exposed a pair of raw C structs from MbedTLS - `mbedtls_x509_crt`  and `mbedtls_pk_context` as a public API. I don't understand why this is necessary? 

I think we better not expose these structures in our public API. They are too low level. We should work in our public API with `MbedTLSX509Crt` and `PkContext` which I renamed to just `Certificate` and `PrivateKey` so that they do mean _something_ to the user rather than the MBedTLS naming gibberish. :)

**The one and only thing remaining** we need to decide is how (the hell :) ) to arrange this bunch of config parameters that the `Session`s take so that they are not too many. That's all.

[Here please find an alternative suggestion](https://github.com/esp-rs/esp-mbedtls/pull/95):
- `Mode` goes down the drain
- `Certificates` goes down the drain
- `SessionConfig` now an enum of either `ServerSessionConfig` or `ClientSessionConfig`
- `ServerSessionConfig` and `ClientSessionConfig` have public members (we can re-assess that) but also constructor and _sensible defaults_ for server vs client
- `Credentials` introduced so that we don't say "when you give us a client certificate, you should also give us a client private key". Either both or none - now enforced

As a result:
- `Session::new` simplified in terms of how many parameters it takes. And you/we can enhance Client/ServerSessionConfig with a builder pattern or whatnot
- edge-nal's `TlsAcceptor`  takes `ServerSessionConfig` rather than a bunch of parameters
- edge-nal's `TlsConnector` takes `ClientSessionconfig` rather than a bunch of parameters

Note that I did not bother updating the examples, yet. As I'm not sure if this is the route to take. 

Thoughts?

PR #78, opened 2025-06-23
Build libraries using Github Actions

AnthonyGrondin:
This PR adds a long-awaited feature; Build libraries using the CI.

When changes occur in `esp-mbedtls-sys/`, the `gen` xtask is triggered to rebuild the libraries, and then tests are executed against the current libraries. When pushing to `main`, the libraries are then committed and pushed to the `main` branch by the action.

Also, when labelling a PR with the `rebuild-libs` label, the CI is programmed to rebuild libraries and push them to the PR branch. When merging to `main`, since no changes will be detected, there won't be a second commit.

This enables rebuilding the libraries without needing to setup the build environment locally on a computer, allows contributors to bring changes to the underlying libraries, while we ensure they are rebuilt in a safe environment, and also allows users to pull back the build libraries to test them locally, if they don't want to set up the environment to build them locally.

Due to the current architecture, executing CI on a push outside of `main` is now disabled. Also, to make the action push the libraries, either the trigger has to be a push to `main`, or adding the label `rebuild-libs` to the PR, since GitHub doesn't support labels on branches directly, so we can't conditionally decide when to push or not, that way.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,10 +133,12 @@ jobs:
     permissions:
       contents: write
     needs: build-test
+    # TODO: Currently GitHub doesn't allow pushing to a forked repo's branch when running an action on a PR to upstream.
     if: |
-      (needs.build-test.outputs.upload-libs &&
+      github.event.pull_request.head.repo.full_name == github.repository &&
+      ((needs.build-test.outputs.upload-libs &&
       github.ref == 'refs/heads/main') ||
-      contains(github.event.pull_request.labels.*.name, 'rebuild-libs')
+      contains(github.event.pull_request.labels.*.name, 'rebuild-libs'))
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,31 @@
-name: CI
+# This CI optionally builds libraries and then builds examples against them.
+#
+# If a change is detected within `esp-mbedtls-sys/`, a rebuild is triggered and this CI will automatically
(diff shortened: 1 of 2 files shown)


AnthonyGrondin:
This seems to need a little tweaking still; It used to work on my fork but when doing PRs across multiple repos, it hits permission issues.

ivmarkov (approved):

bjoernQ:
Anything we can change in regards to the permission problems?

AnthonyGrondin:
> I'm not really very proficient in CI actions, but with that said, I did not notice anything outstanding. Once you resolve the permission issues, I suggest you push it.
> 
> One request I have, but that could definitely wait for the future is to also provide builds for the `thumbv7em-none-eabi` (NRF52XX) and `thumbv6m-none-eabi` (RP2040). This I would need if we are to use `(esp-)mbedtls` into `openthread` which I very much would like to (currently, `openthread` builds and ships its own copy of `mbedtls` which is a bit suboptimal if the user wants to depend on both `esp-mbedtls` _and_ `openthread`, as certain `rs-matter` builds might).
> 
> But this can wait a bit. For that, the CI would need to download and install the ARM GCC toolchain.
> 
> Also, I would soon open a PR which is streamlining a bit the `xtasks`-based build of `esp-mbedtls-sys`, similar to how it is done in `openthread`. There, it is done per target triple, rather than per MCU.

I don't have those boards and I don't have experience in building for them. I believe most of the CI is architecture agnostic so as you said, once we provide the builder for them, we just need the toolchain installed in CI and it should work.

bjoernQ (approved):
I'm not an expert in GitHub CI - LGTM

PR #81, opened 2025-07-15
feat: Implement `core::error:Error` for `TlsError`

### LINKED ISSUE #80: Implement Rust's error trait on TlsError
It would be nice to make `TlsError` implement the `core::error::Error` trait, so that it can more easily be used with other crates like `anyhow`.

AnthonyGrondin:
Closes #80


ivmarkov:
Thanks!

Issue #98, opened 2026-01-15
Build mbedtls `.a` libs without sysroot


ivmarkov:
By "sysroot" we mean the (few?) C headers that `mbedtls` needs to have around when it is being built by the C complier. Like `string.h` but also others.

Also do note, that - even today - we do **not** need the actual `.a` libc during the linking phase from the sysroot. Like `newlib`.
Instead, we are today either linking against a (subset of) libc directly available in the ROM of some MCUs like esp, or we can use the pure-Rust `tinyrlibc`.

So it is really just about the few headers that mbedtls might need so that it gets the signatures of these functions.

Why is the sysroot such a problem and why we might want to get rid of it?
Because to get those C headers for your MCU, you need to **install its GCC cross-compiler**. The host system's sysroot won't do. (Also, `clang` does **not** come with its own sysroot, as it is a cross-compiler by design, so having a sysroot coming with it would mean **multiple** sysroots.)

If we get rid of the sysroot (by defining our own little one - directly in `esp-mbedtls`) we should be able to just use the system-installed `clang` not just to generate the Rust bindings for the C mbedtls lib, but also to compile the C mbedtls lib. And since `clang` is cross-platform and most often than not - installed on the host platform already - suddenly we might no longer need (or it won't be so pressing at least) to have the `.a` artefacts of the C mbedtls pushed into the code repo*. Or require the user to install the GCC cross-toolchain specific to its MCU using whatever obscure ways to do that.

"Just install a clang on your system" would be what we'll ask users.

(*) Unfortunately - for xtensa CPUs specifically - things are not so rosy (yet) because the xtensa arch (esp32/esp32s3/esp32s3) is not yet upstreamed into llvm/clang. But for these chips you currently need a custom Rust toolchain anyway ("esp") which **also** (can) bring a custom clang. so not everything is lost.

EDIT: There's of course the basic problem - with the above - related to types such as `size_t`.
To what is `size_t` equal? The answer is platform-specific. How can we have generic `types.h` file which outputs the correct `typedef`s for the concrete platform?

ivmarkov:
> EDIT: There's of course the basic problem - with the above - related to types such as `size_t`. To what is `size_t` equal? The answer is platform-specific. How can we have generic `types.h` file which outputs the correct `typedef`s for the concrete platform?

Possible approach:
```c
/* Provide types that map to CLANG builtins */
typedef __INT8_TYPE__ int8_t;
[...]

PR #99, opened 2026-01-16
Build with clang by default

ivmarkov:
Addresses #98 

Turns out, inventing a generic, cross-platform "sysroot" for `clang`+`esp-mbedtls` is so simple, that it does not make _any_ sense to compile with GCC by default, and thus ask users to install the (cross) GCC toolchain for their target.

This suddenly makes on-the-fly compilation of esp-mbedtls so much eaiser!

We can just tell users: "install your system clang"*.

(*) 
Caveat 1: System clang (`sudo apt-get install clang`) won't work for xtensa targets until after upstream llvm (and thus clang) get support for the `xtensa` targets. But in the meantime, `espup` would do.
Caveat 2: Users would also need `cmake` + `ninja` pre-installed. For now. But this is also no big deal.

It would have been so nice to have `rustup` distribute `clang` as well, but it seems this effort had died: https://github.com/rust-lang/rust/issues/56371


bjoernQ (approved):
Awesome that we need that little for a suitable mini-sysroot

PR #100, opened 2026-01-16
Restore support for ESP-IDF

ivmarkov:
~~**(NOTE: This needs #99 to be merged first!**~~

Description:

This PR restores support for ESP-IDF.

### Changes to esp-mbedtls / esp-mbedtls-sys

The changes to `esp-mbedtls`/`esp-mbedtls-sys` are just fixing of several imports.

### Changes to the examples

The STD examples were slightly adjusted to work for ESP-IDF which is also a STD platform.

The major change is the introduction of a `bootstrap` module - similar to the one in the ESP baremetal examples - which takes care of ESP-IDF specifics like starting up ESP-IDF logging and connecting to Wifi.

### Changes to CI

CI was enhanced to work for two ESP-IDF targets. Building for all ESP-IDF targets would delay the CI too much, so only 2 are enabled for now.

### Misc

Some comments from @AnthonyGrondin - primarily w.r.t. logging level and cargo flash baudrate from #95 were incorporated here as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -212,7 +212,7 @@ jobs:
         uses: esp-rs/xtensa-toolchain@v1.6
         with:
           default: true
-          ldproxy: false
+          ldproxy: true
 
       # TODO: Double-check and uncomment
       # - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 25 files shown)


bjoernQ (approved):
LGTM

PR #103, opened 2026-01-26
update mbedtls to v3.6.5-idf

ericschaal:
### Upgrade mbedtls to [mbedtls-3.6.5-idf](https://github.com/espressif/mbedtls/tree/mbedtls-3.6.5-idf)

This PR upgrades from 3.4.x to 3.6.5 to resolve TLS 1.3 issues I encountered during testing.

**Context**
- I'm new to both esp-mbedtls and mbedtls in general, so I'm not sure if there was a specific reason for staying on 3.4.x
- The upgrade fixed my TLS 1.3 problems, and basic testing looks good
- However, I haven't done extensive testing across all use cases

Would appreciate review from maintainers familiar with the project to assess whether this is a safe upgrade path.

Addresses https://github.com/esp-rs/esp-mbedtls/issues/72


esp-mbedtls-sys/gen/sysroot/include/inttypes.h
```diff
@@ -0,0 +1,80 @@
+#ifndef __INTTYPES_H__
+#define __INTTYPES_H__
+
+#include <stdint.h>
+
+/* Printf format macros for fixed-width integer types */
+/* Using clang's built-in format string macros for portability */
+
+/* 8-bit integers */
+#define PRId8    __INT8_FMTd__
```
ivmarkov: Can you explain a bit what in the new MbedTLS is using these?

esp-mbedtls-sys/gen/sysroot/include/inttypes.h
```diff
@@ -0,0 +1,80 @@
+#ifndef __INTTYPES_H__
+#define __INTTYPES_H__
+
+#include <stdint.h>
```
ivmarkov: Hmmmm, from where is this coming?
The whole idea with `clang` was that _we_ are providing a sysroot and not relying on any headers to be around.

ivmarkov: Hmmm, we are already having that issue with e.g. `stddef.h`, which is included from... somewhere...

ivmarkov:
@ericschaal I think the PR looks good! If only you could:
- Restore the imports' order (there is a comment about that)
- Explain a bit why we have to introduce `inttypes.h` with so many types inside

ivmarkov:
> I think we should also update config.h to the latest version as done in [this commit](https://github.com/ericschaal/esp-mbedtls/commit/f1220b5f6046994638fe60389dd9ba4a55d2a5f1).

@ericschaal Good idea. Can you pull that update into this PR?

ivmarkov:
Thanks! 

Tested and works on esp32s3, esp32c3 and esp32c6.
Also works with `rs-matter`.

PR #106, opened 2026-02-04
Fix memory alignment issues due to mbedtls_xxx types being moveable

ivmarkov:
**TL;DR: This is a relatively urgent fix as the issue it addresses might result into undefined behavior (dirty crashes with memory corruption).**

Long story:
The way we hook algos like sha1, sha256 etc. "in Rust" without requiring re-compilation of the MbedTLS code (and worse - `CBindGen` usage) for each possible HW-accelerated hook the user might place, is that we define, e.g. the "Sha256" state **[as a bunch of bytes](https://github.com/esp-rs/esp-mbedtls/blob/8d76b85bcee70891c45b3e6afafe51eef9ec72bd/esp-mbedtls-sys/gen/include/sha256_alt.h#L2)**. Where this bunch of bytes is "big enough" (arbitrary decision based on gut feeling) so that any driver should fit (there will be overrides/features for that in future).

It is then up to the user and the HW driver it hooks to re-interpret these bytes as the actual type `T` of the driver.

The idea was copied from `openthread` and relies on Rust's `core::slice::align_to_mut` to operate.

So far so good, **except** (and unlike the case in `openthread`!) these bunch of bytes [**can move**](https://github.com/sysgrok/rs-matter/blob/c5cf7f0563a57ffb241a153ce7023ce8b2d6a99c/rs-matter/src/crypto/backend/mbedtls.rs#L309). They are not pinned or boxed to a memory location.

What this means is that the **position** (in memory) of this bunch of bytes defines **how exactly** the type `T` is aligned inside the slice of bytes when using `core::slice::align_to_mut`. For example, if the bunch of bytes starts at an address which is only divisible by 2 but not by 4 - yet - `T` has an alignment requirement = 4, then `T` will start at offset 2 instead of at offset 0 in the bunch of bytes.

So... if the bunch of bytes moves (or gets copied), suddenly, `core::slice::align_to_mut` might end up with a different calculation for the offset of `T` inside these bytes!

The solution/workaround:
- Change the "bunch of bytes" ~~into a "bunch of `u64`s"~~ to have a custom alignment of 16.
- Strictly speaking, the custom alignment should be (in C speak) `__attribute__((aligned(_Alignof(__int128))))`, but I can't make `clang` understand `_Alignof(__int128)` likely because `__int128` is a weird nonstandard type to begin with. The reason why we want the latter, is because alignment of `u128` on riscv32 is 8 instead of 16, so these platforms do not have to necessarily pay the price of full alignment. But alas, it is what it is.

This way we guarantee that the bunch of bytes will be aligned at a 16-byte boundary. Which - **in turn** - guarantees that any type `T` which itself has alignment requirement <= 16 bytes - will always be emplaced at the beginning of the bunch of bytes, because these would already be aligned at a 16 bytes boundary by the C compiler.

~~Going with `u128` seemed a bit over the top for me so yeah - don't use `u128`s in your hooks. Or use them on HW architectures, where the alignment of u128 is 8 instead of 16 bytes. (grin)~~ EDIT: Since RustCrypto Sha512 wants alignment of 16, I did it.


bjoernQ (approved):
Thanks

PR #116, opened 2026-03-16
Add suppor for session reuse

yanshay:
I added session reuse - it is required for certain ftp servers that won't work without it, will reject.
Initially it didn't work and I couldn't figure out why. Five minutes after submitting this PR asking for help I found the root cause.

The code I added is code that worked on a previous version of esp-mbedtls right when esp-hal 1.0.0 came out (I used PR96), adapted to the new conventions introduced since then.

The change in the new version that prevents this from working, is that `connect()` begins with :

```rust
    /// Establish the SSL connection
    async fn connect(&mut self) -> Result<(), SessionError> {
        debug!("Establishing SSL connection");

     merr!(unsafe { mbedtls_ssl_session_reset(self.ssl_context as *const _ as *mut _) })?;

```

`mbedtls_ssl_session_reset` resets the context right after my connect_reuse fills in the session to reuse in the context, so obviously it can't reuse the session.

I remarked this line in my code and it is now working, including the connect that isn't reused.

1. is session_reset required at the beginning of connect? in previous version it was called only in case of certain errors:

```rust
            loop {
                let res = mbedtls_ssl_handshake(self.ssl_context);
                log::debug!("mbedtls_ssl_handshake: {res:x}");
                if res == 0 {
                    // success
                    break;
                }
                if res < 0 && res != MBEDTLS_ERR_SSL_WANT_READ && res != MBEDTLS_ERR_SSL_WANT_WRITE
                    // See https://github.com/Mbed-TLS/mbedtls/issues/8749
                    && res != MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET
                {
                    // real error
                    // Reference: https://os.mbed.com/teams/sandbox/code/mbedtls/docs/tip/ssl_8h.html#a4a37e497cd08c896870a42b1b618186e
                    mbedtls_ssl_session_reset(self.ssl_context);
                    #[allow(non_snake_case)]
                    return Err(match res {
                        MBEDTLS_ERR_SSL_NO_CLIENT_CERTIFICATE => TlsError::NoClientCertificate,
                        _ => TlsError::MbedTlsError(res),
                    });
                }

                // try again immediately
            }

```

2. If it is required, how would you recommend I solve this?

### DIFF
--- a/mbedtls-rs/src/session.rs
+++ b/mbedtls-rs/src/session.rs
@@ -11,7 +11,7 @@ mod asynch;
 pub mod blocking;
 
 /// A reusable TLS session state captured from a connected session.
-pub struct ReusedSession {
+pub struct SavedSession {
     pub(crate) mbedtls_session: MBox<mbedtls_ssl_session>,
 }
 
@@ -279,10 +279,6 @@ pub enum SessionError {
     MbedTls(MbedtlsError),
     /// IO error
     Io(ErrorKind),
-    /// Session is already connected
-    AlreadyConnected,
-    /// Session is not connected
-    NotConnected,
 }
 
 impl SessionError {
@@ -309,8 +305,6 @@ impl core::fmt::Display for SessionError {
         match self {
             Self::MbedTls(e) => write!(f, "{}", e),
             Self::Io(e) => write!(f, "IO({:?})", e),
-            Self::AlreadyConnected => write!(f, "Session already connected"),
-            Self::NotConnected => write!(f, "Session not connected"),
         }
     }
 }
--- a/mbedtls-rs/src/session.rs
+++ b/mbedtls-rs/src/session.rs
@@ -10,6 +10,11 @@ pub use asynch::*;
 mod asynch;
 pub mod blocking;
 
+/// A reusable TLS session state captured from a connected session.
+pub struct ReusedSession {
+    pub(crate) mbedtls_session: MBox<mbedtls_ssl_session>,
+}
+
 /// Certificate verification mode used for a session
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
@@ -274,6 +279,10 @@ pub enum SessionError {
     MbedTls(MbedtlsError),
     /// IO error
     Io(ErrorKind),
+    /// Session is already connected
+    AlreadyConnected,
+    /// Session is not connected
+    NotConnected,
 }
 
 impl SessionError {
@@ -300,6 +309,8 @@ impl core::fmt::Display for SessionError {
         match self {
             Self::MbedTls(e) => write!(f, "{}", e),
             Self::Io(e) => write!(f, "IO({:?})", e),
+            Self::AlreadyConnected => write!(f, "Session already connected"),
+            Self::NotConnected => write!(f, "Session not connected"),
         }
     }
 }
@@ -310,6 +321,8 @@ impl defmt::Format for SessionError {
         match self {
             Self::MbedTls(e) => defmt::write!(f, "{}", e),
             Self::Io(e) => defmt::write!(f, "IO({:?})", debug2format!(e)),
+            Self::AlreadyConnected => defmt::write!(f, "Session already connected"),
+            Self::NotConnected => defmt::write!(f, "Session not connected"),
         }
     }
 }
--- a/mbedtls-rs/src/session.rs
+++ b/mbedtls-rs/src/session.rs
@@ -315,8 +315,6 @@ impl defmt::Format for SessionError {
         match self {
             Self::MbedTls(e) => defmt::write!(f, "{}", e),
             Self::Io(e) => defmt::write!(f, "IO({:?})", debug2format!(e)),
-            Self::AlreadyConnected => defmt::write!(f, "Session already connected"),
-            Self::NotConnected => defmt::write!(f, "Session not connected"),
         }
     }
 }
--- a/mbedtls-rs/src/session/asynch.rs
+++ b/mbedtls-rs/src/session/asynch.rs
@@ -11,7 +11,7 @@ use crate::sys::MbedtlsError;
 use crate::sys::*;
 use crate::{SessionError, TlsReference};
 
-use super::{ReusedSession, SessionConfig, SessionState};
+use super::{SavedSession, SessionConfig, SessionState};
 
 /// Re-export of the `embedded-io-async` crate so that users don't have to explicitly depend on it
 /// to use e.g. `write_all` or `read_exact`.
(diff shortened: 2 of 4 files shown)


ivmarkov:
EDIT: Removed the `MRc` in favor of your original `MBox` design, because - as it turns out - the saved session - when loaded - is only referenced during loading, then can be discarded.

> 2. If it is required, how would you recommend I solve this?

How about (in **both** `blocking` and `asynch`):
- Rename the existing `connect` to `connect_internal` and make it private
- Introduce a new parameter to `connect_internal` => `saved_session: Option<&SavedSession>`
- Re-introduce `connect` as a one-liner to `connect_internal(None)`
- Introduce `reconnect(saved_session: SavedSession)` as a one-liner to `connect_internal(Some(saved_session))`

As for the `SavedSession` type, it should be something like (as you actually did already):
```rust
pub struct SavedSession {
      raw: MBox<mbedtls_ssl_session>,
}
```

And then both the blocking and async sessons should have an extra method:
```rust
fn save(&mut self) -> Result<SavedSession, MbedtlsError> { ... }
```

Please do **not** extend `SessionError` with your own error variants. Stick to the `MbedtlsError` error codes.

ivmarkov:
Oh, almost forgot. In `connect_internal` just call `mbedtls_ssl_set_session` **AFTER** the reset call. Obviously, `mbedtls_ssl_set_session` needs to be called only if the `Option` containing the saved session is "some".

mbedtls-rs/src/session/asynch.rs
```diff
@@ -124,6 +125,27 @@ where
         Ok(())
     }
 
+    /// Negotiate the TLS connection attempting to reuse a previously captured session.
+    pub async fn connect_reuse(
```
ivmarkov: See [my comment](https://github.com/esp-rs/mbedtls-rs/pull/116#issuecomment-4089092113) as to how to arrange the various `connect` methods.

mbedtls-rs/src/session/asynch.rs
```diff
@@ -225,6 +247,26 @@ where
 
         Ok(())
     }
+
+    /// Capture the negotiated MbedTLS session for possible reuse.
+    pub fn get_mbedtls_session(&self) -> Result<ReusedSession, SessionError> {
```
ivmarkov: As per [my comment[(https://github.com/esp-rs/mbedtls-rs/pull/116#issuecomment-4089092113), rename this to `save`

mbedtls-rs/src/session/asynch.rs
```diff
@@ -225,6 +247,26 @@ where
...
+        if !self.connected {
+            return Err(SessionError::NotConnected);
```
ivmarkov: As per [my comment[(https://github.com/esp-rs/mbedtls-rs/pull/116#issuecomment-4089092113), avoid introducing new variants of SessionError

mbedtls-rs/src/session.rs
```diff
@@ -10,6 +10,11 @@ pub use asynch::*;
 mod asynch;
 pub mod blocking;
 
+/// A reusable TLS session state captured from a connected session.
+pub struct ReusedSession {
```
ivmarkov: As per [my comment[(https://github.com/esp-rs/mbedtls-rs/pull/116#issuecomment-4089092113), rename to `SavedSession`

mbedtls-rs/src/session.rs
```diff
@@ -274,6 +279,10 @@ pub enum SessionError {
     MbedTls(MbedtlsError),
     /// IO error
     Io(ErrorKind),
+    /// Session is already connected
```
ivmarkov: Do we absolutely need these? What would `mbedtls_ssl_get_session` do if calling it when the session is not connected? Can't we just call it unconditionally?

As for "AlreadyConnected", also not necessary. Re-connect should disconnect and connect again, if so the user has requested.

ivmarkov (changes requested):

ivmarkov:
@yanshay Are you planning to push this to completion? This is the one thing stopping me from publishing `mbedtls-rs` on crates.io

yanshay:
Didn't know that was in plan.

Will get to it in the next few days.

yanshay:
I'm good with your suggestions and only not sure about the function 'reconnect' name.

The use case I'm aware of is opening a new connection on a new socket/port so it's not a reconnect after some connection was closed, that's why in my case I used connect_reuse.

Are there additional cases reusing an existing session is used? Or is it possible to really reconnect after disconnect using this?
Maybe call is connect_with_session if connect_reuse is just one use case for this?

As for errors maybe I will just perform the operations anyway and return the mbedtls errors.

ivmarkov:
> I'm good with your suggestions and only not sure about the function 'reconnect' name.
> 
> The use case I'm aware of is opening a new connection on a new socket/port so it's not a reconnect after some connection was closed, that's why in my case I used connect_reuse.
> Are there additional cases reusing an existing session is used? Or is it possible to really reconnect after disconnect using this? Maybe call is connect_with_session if connect_reuse is just one use case for this?

Agreed: let's use `connect_reuse`.
I admit that `reconnect` does somehow imply reconnecting an already existing socket. I reached for a shorter name, but it might be misleading indeed.

> 
> As for errors maybe I will just perform the operations anyway and return the mbedtls errors.

Yes, if possible, that would be great!

yanshay:
> Agreed: let's use connect_reuse.

Checked with ChatGPT about naming, it suggested connect_with_session so went with it, ok?

ivmarkov:
> > Agreed: let's use connect_reuse.
> 
> Checked with ChatGPT about naming, it suggested connect_with_session so went with it, ok?

np.

mbedtls-rs/src/session/asynch.rs
```diff
@@ -7,10 +7,11 @@ use embedded_io::ErrorKind;
 
 use io::{ErrorType, Read, Write};
 
+use crate::sys::MbedtlsError;
```
ivmarkov: Nit: we glob-import `sys::*` just below, so why do we import`MbedtlsError`?

mbedtls-rs/src/session.rs
```diff
@@ -310,6 +315,8 @@ impl defmt::Format for SessionError {
         match self {
             Self::MbedTls(e) => defmt::write!(f, "{}", e),
             Self::Io(e) => defmt::write!(f, "IO({:?})", debug2format!(e)),
+            Self::AlreadyConnected => defmt::write!(f, "Session already connected"),
```
ivmarkov: This we should delete?

mbedtls-rs/src/session.rs
```diff
@@ -310,6 +315,8 @@ impl defmt::Format for SessionError {
...
+            Self::NotConnected => defmt::write!(f, "Session not connected"),
```
ivmarkov: ditto?

ivmarkov (commented):
@yanshay Looks quite OK!
Just two small comments:
- A minor import nit
- If you could remove the new enum variants from `SessionError`; you should no longer be using those, no?

Once you do these, and if you have tested that session reuse works (at least the async variant, blocking is even simpler), then let me know and I'll merge.

ivmarkov (approved):

ivmarkov:
Thanks!

yanshay:
Thank you for all the work you do on this important crate!

PR #117, opened 2026-03-19
Re-arrange the modules for the various hook impls

ivmarkov:
As discussed [here](https://github.com/esp-rs/mbedtls-rs/pull/108#issuecomment-4088548067), this is a follow-up to #108 that aims to:
- Reduce the number of features exposed by `mbedtls-rs(-sys)`
- Get them hook impls under a common module, so that they are not everywhere in the sys crate, left and right

Note that while it was possible to reduce all `esp32XX` features into a single one - `esp` or `esp-hal` via introducing a build-time dependency on `esp-metadata-generated`, I decided NOT to do it for now, as I'm not sure the benefits will outweight the disadvantage that the user should depend on the **exact** `esp-metadata-generated` version used in `mbedtls-rs-sys`. See the whole discussion here: https://matrix.to/#/!LdaNPfUfvefOLewEIM:matrix.org/$EONcLSsC_F-_3loqLUP3KRPmda1z8mFI53y3mQQNXKs?via=matrix.org&via=mozilla.org&via=tchncs.de


mbedtls-rs-sys/src/hook/backend.rs
```diff
@@ -0,0 +1,6 @@
+#[cfg(feature = "embassy-time")]
+pub mod embassy;
+#[cfg(feature = "esp-hal")]
```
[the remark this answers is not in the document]
ivmarkov: This is bullshit, because the esp32XX features each enable the `esp-hal` dependency, which is also a feature.

ivmarkov:
@ericschaal If you would like to take a look?

ericschaal:
@ivmarkov taking a look asap!

mbedtls-rs-sys/Cargo.toml
```diff
@@ -19,37 +19,28 @@ use-gcc = []
...
-wall-clock-std = ["time", "hook-wall-clock"]
-wall-clock-esp32 = ["esp-hal/esp32", "time", "hook-wall-clock"]
-wall-clock-esp32c2 = ["esp-hal/esp32c2", "time", "hook-wall-clock"]
-wall-clock-esp32c3 = ["esp-hal/esp32c3", "time", "hook-wall-clock"]
-wall-clock-esp32c6 = ["esp-hal/esp32c6", "time", "hook-wall-clock"]
-wall-clock-esp32h2 = ["esp-hal/esp32h2", "time", "hook-wall-clock"]
-wall-clock-esp32s2 = ["esp-hal/esp32s2", "time", "hook-wall-clock"]
-wall-clock-esp32s3 = ["esp-hal/esp32s3", "time", "hook-wall-clock"]
+# esp-hal hook impls
+esp32 = ["esp-hal/esp32", "crypto-bigint", "time"]
```
ericschaal, line 37: FYI `time` is only needed when the `hook-wall-clock` feature is enabled. 

On esp* when `hook-wall-clock` is enabled, `EspRtcWallClock` is compiled in, requiring the `time` crate.
This comment also applied to `std = ["time"]` above as the mecanics are about the same but for `StdWallClock`

ericschaal:
Changes make sense to me, I guess the only flexibility that was lost is allowing users to opt-in timer/wall-clock without compiling the default implementations. But I don't think it matters as users can just decide not to use the default impl and implement their own and hook them using `hook_wall_clock` and `hook_timer`  directly.

mbedtls-rs-sys/Cargo.toml
ivmarkov: I get that, but it is unclear to me how exactly we can express that using the declarative way of cargo. And I doubt this is possible. Hence why `time` is enabled when `esp` support is enabled, even if wall-clock hooking is not on.

ivmarkov:
> Changes make sense to me, I guess the only flexibility that was lost is allowing users to opt-in timer/wall-clock without compiling the default implementations. 

The default implementations will not be available unless - in addition to enabling the timer/wall-clock hooks - the user enabled the `esp32XX` features(s) or the STD features.

> But I don't think it matters as users can just decide not to use the default impl and implement their own and hook them using `hook_wall_clock` and `hook_timer` directly.

Yes. and if the default impls are not used, they should be trimmed out at link time during the dead code removal phase.

mbedtls-rs-sys/Cargo.toml
ericschaal: Fair, I was thinking of adding it to hook-wall-clock:
```
hook-wall-clock = ["hook-timer", "dep:time"]
```
but it feels awkward since time is a backend dependency and has nothing to do with the hook definition itself.

PR #119, opened 2026-03-30
Include directories metadata also for pre-built targets; prep for crates.io release

ivmarkov:
The PR is addressing three issues:

##### Issue 1 - with `cargo::metadata=include`

There is a long-standing TODO there, that this key is emitted only when an on-the-fly build is being performed.
However, we need to emit this key also when the pre-built `.a` libraries are used. Downstream crates wrapping C libs - like the `openthread` crate _might_ actually end up using the pre-built `.a` libraries, but then they would _still_ need the mbedtls .h files.

##### Issue 2 - prep for crates.io release

For releasing to crates.io, we must make sure that we are grabbing everything necessary in the about-to-be-published package:
- The whole Rust source code (obviously)
- But also, the C source-code of the `mbedtls` submodule. A relatively minimal subset of it that allows the mbedtls C lib to be built on the fly also when consuming `mbedtls-rs` from `crates.io`

##### Issue 3 - the mbedtls `.a` artefacts were "installed" at the end of the build

We don't really need the installation step even if the installation prefix can be controlled by us and pointed at a safe directory. Also this step creates a problem in `rs-matter`'s CI.

### DIFF
--- a/mbedtls-rs/Cargo.toml
+++ b/mbedtls-rs/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 rust-version = "1.84"
-description = "Rust library implementing transparent TLS encryption/decrypton for IO streams with MbedTLS"
+description = "Rust library implementing transparent TLS encryption/decryption for IO streams with MbedTLS"
 repository = "https://github.com/esp-rs/mbedtls-rs"
 readme = "README.md"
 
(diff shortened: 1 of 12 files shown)


mbedtls-rs/Cargo.toml
```diff
@@ -4,6 +4,9 @@ version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 rust-version = "1.84"
+description = "Rust library implementing transparent TLS encryption/decrypton for IO streams with MbedTLS"
```
bjoernQ: typo "decrypton"

bjoernQ (approved):
LGTM

mbedtls-rs/Cargo.toml
ivmarkov: Thanks - fixed there and also in the README, from where the typo was copy-pasted.

PR #124, opened 2026-04-24
Add support for ALPN

anszom:
This PR adds support for the ALPN feature of TLS.

Controversial topics:

- `ALPNArray` is used internally to create (via mbedtls_calloc) a null-terminated pointer array from a `[&CStr]` slice. An alternative approach would be to have the client create the storage on their own, comparable to CStr & CString.
- extending *SessionConfig is a breaking API change. I couldn't find any good way around this.

### DIFF
--- a/mbedtls-rs/src/session/asynch.rs
+++ b/mbedtls-rs/src/session/asynch.rs
@@ -88,6 +88,20 @@ where
         unsafe { mbedtls_ssl_get_verify_result(&*self.state.ssl_context) }
     }
 
+    /// Get the negotiated ALPN protocol, if any.
+    ///
+    /// NOTE: This function should be called only after a `connect()` call.
+    pub fn tls_alpn(&self) -> Option<&CStr> {
+        unsafe {
+            let ptr = mbedtls_ssl_get_alpn_protocol(&*self.state.ssl_context);
+            if ptr.is_null() {
+                None
+            } else {
+                Some(CStr::from_ptr(ptr))
+            }
+        }
+    }
+
     /// Get a mutable reference to the underlying stream
     pub fn stream(&mut self) -> &mut T {
         &mut self.stream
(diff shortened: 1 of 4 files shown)


anszom:
I can't see how my changes would be related to the CI failures...

mbedtls-rs/src/session/asynch.rs
```diff
@@ -88,6 +88,20 @@ where
         unsafe { mbedtls_ssl_get_verify_result(&*self.state.ssl_context) }
     }
 
+    /// Get the negotiated ALPN protocol, if any.
+    ///
+    /// NOTE: This function should be called only after a `connect()` call.
+    pub fn tls_alpn(&'a self) -> Option<&'a [u8]> {
```
ivmarkov: - Why `&[u8]`? This is a `CStr`
- `&'a` is redundant and once clippy clicks-in it will complain about that.

mbedtls-rs/src/session/blocking.rs
```diff
@@ -143,6 +143,20 @@ where
         unsafe { mbedtls_ssl_get_verify_result(&*self.state.ssl_context) }
     }
 
+    /// Get the negotiated ALPN protocol, if any.
+    ///
+    /// NOTE: This function should be called only after a `connect()` call.
+    pub fn tls_alpn(&'a self) -> Option<&'a [u8]> {
```
ivmarkov: ditto.

ivmarkov (commented):
Just two nits, if you could address.

As for the Alpn NULL-terminated array being controversial... it is but I can't come up with a significantly better idea either.

anszom:
I've only just noticed that esp-idf builds were failing. Fixed that now. To avoid repetition, I've promoted mbedtls_calloc & free to pub(crate)

PR #127, opened 2026-05-30
fix(build): make CMake config work with multi-config generators

andrico21:
## Summary

`gen/builder.rs::CMakeConfigurer::configure_base` sets `CMAKE_BUILD_TYPE=MinSizeRel`, but:

1. The `cmake` crate's `Config::build()` later forces `--config Release` (and we use it via `.profile("Release")` in `compile()`), so the build type set here is ignored regardless.
2. Multi-config generators (Ninja Multi-Config, Visual Studio, Xcode) ignore `CMAKE_BUILD_TYPE` entirely and also ignore the unsuffixed `CMAKE_*_OUTPUT_DIRECTORY` vars. They place artifacts under `<dir>/<Config>/`, so the existing `CMAKE_ARCHIVE_OUTPUT_DIRECTORY=<target>` has no effect — link step then fails to find `libmbedtls.a` / `libmbedcrypto.a` / `libmbedx509.a` under the expected path.

## Change

- Drop the dead `CMAKE_BUILD_TYPE=MinSizeRel` define.
- When `target_dir` is provided, add three multi-config-aware defines so the same flat layout is used regardless of generator:
  - `CMAKE_CONFIGURATION_TYPES=Release` (restrict generated configs)
  - `CMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELEASE=<target>`
  - `CMAKE_LIBRARY_OUTPUT_DIRECTORY_RELEASE=<target>`
  - `CMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE=<target>`

Net effect: single-config generators (Make, plain Ninja) continue to use the existing unsuffixed `*_OUTPUT_DIRECTORY` vars; multi-config generators now honour the per-config equivalents and produce the same flat `<target_dir>/lib*.a` layout. Aligned with the `.profile("Release")` already used by the call site.

## Why this matters

On Windows with the bundled `cmake` 4.x (e.g. `pip install cmake` → `cmake.exe` in venv `Scripts/`), CMake picks `Ninja Multi-Config` by default if available, so this bites downstream users who don't override `CMAKE_GENERATOR=Ninja` explicitly. Without this fix the build script panics with `cmake-rs` saying `did not execute successfully` after `cmake --build` succeeds but the `.a` files land under `<target>/Release/` rather than `<target>/`.

## Validation

Tested with three configurations against this branch (rebased on top of current `main`):

| Generator | Output path before fix | After fix |
|---|---|---|
| Ninja (single-config) | `<target>/lib*.a` ✅ | `<target>/lib*.a` ✅ |
| Ninja Multi-Config | `<target>/Release/lib*.a` ❌ (link fails) | `<target>/lib*.a` ✅ |
| Visual Studio 17 2022 | `<target>/Release/lib*.a` ❌ (link fails) | `<target>/lib*.a` ✅ |

Also confirmed end-to-end build via the `examples-esp` crate (rebased branch) for `esp32c3`, `esp32c6`, and `esp32c5` targets.

## Scope

Single-file 4-line change in `mbedtls-rs-sys/gen/builder.rs`. No public API or feature changes.

### DIFF
--- a/mbedtls-rs-sys/gen/builder.rs
+++ b/mbedtls-rs-sys/gen/builder.rs
@@ -367,7 +367,7 @@ impl MbedtlsBuilder {
             .define("MBEDTLS_FATAL_WARNINGS", "OFF")
             .define("MBEDTLS_USER_CONFIG_FILE", user_config_path)
             .cflag(format!("-I{}", hook_header_dir.display()))
-            .profile("Release")
+            .profile("MinSizeRel")
             .out_dir(&target_dir);
 
         config.build();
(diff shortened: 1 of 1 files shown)


ivmarkov (commented):
We really DO want MinSizeRel by default.
If - for some reason - this is not working, then this should rather be fixed rather than "fixing" it by switching it to Release.

andrico21:
Good catch. Restored `CMAKE_BUILD_TYPE = MinSizeRel` and also switched `.profile("Release")` → `.profile("MinSizeRel")` at line 370 — without that, the cmake-crate's `cmake --build --config Release` was already overriding `MinSizeRel` on single-config generators, so the original intent was silently lost regardless of generator. Multi-config vars updated to `_MINSIZEREL` to match.

Verified on `Ninja` and `Ninja Multi-Config` — CMakeCache shows `CMAKE_BUILD_TYPE=MinSizeRel`, all 5 libs produced.

Issue #123, opened 2026-04-23
Cover building with GCC in CI


AnthonyGrondin:
PR #108 recently introduced a compilation error when building with GCC:

```rust
error[E0425]: cannot find type `__int_least64_t` in this scope
    --> esp-mbedtls/examples/esp/target/xtensa-esp32s3-none-elf/debug/build/mbedtls-rs-sys-07cc9d2287b97e20/out/bindings.rs:1373:19
     |
1373 | pub type time_t = __int_least64_t;
     |                   ^^^^^^^^^^^^^^^ not found in this scope
```

Since Clang became forced by default in #99 (a9e807a), the CI didn't catch this breakage for users who still rely on GCC. 

To prevent regressions like this in the future, it would be great to add a GCC build job to the CI matrix alongside Clang.

ivmarkov:
Addressed by #179

PR #128, opened 2026-05-30
fix(build): fall back to -print-search-dirs when GCC -print-sysroot is empty

andrico21:
## Summary

`gen/builder.rs::CMakeConfigurer::derive_sysroot` calls `<compiler> -print-sysroot` to locate libc headers for bindgen (passed to bindgen as `--sysroot=<dir>` so it can find `stdio.h`, `stdint.h`, etc.). Some packaged GCC cross-toolchains print an **empty string** for this option even though the sysroot is fully present on disk:

- PlatformIO's `toolchain-riscv32-esp` (Espressif crosstool-NG, GCC 8.4.0, `esp-2021r2-patch5`)
- PlatformIO's `toolchain-xtensa-esp32s3` (same GCC vintage)

These ship with the sysroot under `<prefix>/<triple>/include/` and `<prefix>/<triple>/lib/`, but `gcc -print-sysroot` returns the empty string because crosstool-NG didn't configure `--with-sysroot=<dir>` when building the toolchain. The result: `derive_sysroot()` returns `None`, bindgen is invoked without `--sysroot`, and the build fails with `'stdio.h' file not found`.

## Change

After the existing `-print-sysroot` probe, when the result is empty, fall back to `<compiler> -print-search-dirs` and parse the `install:` line. The path it reports is `<prefix>/lib/gcc/<triple>/<version>/`. Walk four `parent()` calls to reach `<prefix>`, then join the target triple (parsed from the compiler binary name by stripping `.exe` and `-gcc`).

Guard the fallback with a `<prefix>/<triple>/include/stdio.h` existence check so we never return a bogus path: if the layout doesn't match (e.g. a stripped distribution), we keep returning `None` instead of fabricating a wrong sysroot.

## Validation

Tested with three toolchains:

| Toolchain | GCC | `-print-sysroot` | After fix |
|---|---|---|---|
| PlatformIO `toolchain-riscv32-esp` | 8.4.0 | empty | ✅ resolves via fallback |
| `espup`-installed `riscv32-esp-elf-gcc` | 13.2.0 | `<prefix>/riscv32-esp-elf` | ✅ unchanged (returns from `-print-sysroot`) |
| Vanilla `riscv64-linux-gnu-gcc` (Debian) | 12.2.0 | `/usr/riscv64-linux-gnu` | ✅ unchanged (returns from `-print-sysroot`) |

End-to-end build of `examples-esp` against PlatformIO toolchain now succeeds for esp32c3, esp32c6, and esp32c5 targets where it previously failed at bindgen.

## Backward compatibility

- `-print-sysroot` path unchanged: when the toolchain reports a non-empty sysroot we return immediately, identical to current behaviour.
- The fallback only runs when `-print-sysroot` returns an empty string AND status is success.
- Existence check on `stdio.h` ensures we never silently return a directory that doesn't contain headers.

## Scope

Single-file change in `mbedtls-rs-sys/gen/builder.rs`, +43/-4 lines, no public API or feature changes.

## Related

Issue #123 reports the same `stdio.h not found` symptom against PlatformIO toolchain.

### DIFF
--- a/mbedtls-rs-sys/gen/builder.rs
+++ b/mbedtls-rs-sys/gen/builder.rs
@@ -538,15 +538,58 @@ impl CMakeConfigurer {
         let (compiler, gnu) = unforce_clang.derive_c_compiler();
 
         if gnu {
-            let output = Command::new(compiler).arg("-print-sysroot").output().ok()?;
+            let output = Command::new(&compiler)
+                .arg("-print-sysroot")
+                .output()
+                .ok()?;
 
             if output.status.success() {
                 let sysroot = String::from_utf8(output.stdout).ok()?.trim().to_string();
 
-                (!sysroot.is_empty()).then_some(PathBuf::from(sysroot))
-            } else {
-                None
+                if !sysroot.is_empty() {
+                    return Some(PathBuf::from(sysroot));
+                }
             }
+
+            // Some packaged GCC cross-toolchains (e.g. PlatformIO's
+            // `toolchain-riscv32-esp` / `toolchain-xtensa-esp32s3`, both based
+            // on crosstool-NG esp-2021r2-patch5 / GCC 8.4.0) print an empty
+            // string for `-print-sysroot` even though the sysroot is present
+            // on disk. Fall back to deriving the prefix from
+            // `-print-search-dirs` (the `install:` line points at
+            // `<prefix>/lib/gcc/<triple>/<version>/`) and then joining the
+            // target triple.
+            let install_dir = Command::new(compiler)
+                .arg("-print-search-dirs")
+                .output()
+                .ok()
+                .and_then(|o| String::from_utf8(o.stdout).ok())
+                .and_then(|s| {
+                    s.lines()
+                        .find(|l| l.starts_with("install:"))
+                        .map(|l| PathBuf::from(l.trim_start_matches("install:").trim()))
+                });
+
+            if let Some(install_dir) = install_dir {
+                // install_dir = <prefix>/lib/gcc/<triple>/<version>/
+                // Walk up 4 levels (version -> triple -> gcc -> lib) to recover <prefix>.
+                if let Some(prefix) = install_dir
+                    .parent() // <version>
+                    .and_then(|p| p.parent()) // <triple>
+                    .and_then(|p| p.parent()) // gcc
+                    .and_then(|p| p.parent())
+                // lib  ->  <prefix>
+                {
+                    if let Some(triple) = self.derive_gcc_target_triple() {
+                        let candidate = prefix.join(triple);
+                        if candidate.join("include").join("stdio.h").exists() {
+                            return Some(candidate);
+                        }
+                    }
+                }
+            }
+
+            None
         } else {
             None
         }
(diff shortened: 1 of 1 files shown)


ivmarkov:
@andrico21 Pass this to your LLM:

======

# PR #128 — GCC sysroot fallback

## What it does

`derive_sysroot()` in [mbedtls-rs-sys/gen/builder.rs](https://github.com/esp-rs/mbedtls-rs/pull/128/files) supplies the path that later gets fed to bindgen as `--sysroot=...` and `-I<sysroot>/include`. For GCC toolchains it normally runs `gcc -print-sysroot`. On certain ESP toolchains shipped via PlatformIO (`toolchain-riscv32-esp`, `toolchain-xtensa-esp32s3`, GCC 8.4.0), that command exits 0 but prints an empty string, so bindgen can't find `stdio.h` and the build dies.

The PR adds a fallback: if `-print-sysroot` is empty, parse `gcc -print-search-dirs`, find the `install:` line (which points to `<prefix>/lib/gcc/<triple>/<version>/`), walk four parents up to `<prefix>`, then join the target triple recovered from the compiler binary's filename (`<triple>-gcc[.exe]`). Sanity-check by probing for `<candidate>/include/stdio.h`; only return on hit.

## Assessment

**The approach is reasonable.** Empty `-print-sysroot` from GCC really does happen in the wild — typically when the toolchain was relocated and the configured-in sysroot path no longer matches reality, or when the toolchain wasn't built with `--with-sysroot=` at all. Falling back to `-print-search-dirs` is the right second source: it's resolved at runtime from the actual binary location, so it survives the toolchain being unpacked to an arbitrary prefix (which is exactly the PlatformIO case).

**The `stdio.h` probe is the strongest part.** Returning a wrong sysroot here would make bindgen fail later with a more confusing error. Validating the candidate before returning means a bad guess just falls through to `None`, and any existing downstream logic for "no sysroot" kicks in. Good defensive design.

**A few real concerns:**

1. **The result shape differs from the primary path.** The `-print-sysroot` branch returns whatever GCC printed (typically `<prefix>/<triple>` for cross toolchains, but sometimes `<prefix>/<triple>/sysroot` or `<prefix>` itself, depending on how GCC was configured). The fallback always returns `<prefix>/<triple>`. For the targeted ESP toolchains that's correct because their headers live at `<prefix>/<triple>/include/stdio.h` — which is exactly what the `stdio.h` probe verifies. So it's safe *for the cases it accepts*, but the two branches aren't symmetric, and a future toolchain that has a real `sysroot/` subdirectory but an empty `-print-sysroot` would not be discovered. Probably fine; worth a comment.

2. **Triple recovery from the binary filename is brittle.**
   ```rust
   let name = compiler.file_name().unwrap_or_default().to_string_lossy();
   let triple = name.trim_end_matches(".exe").trim_end_matches("-gcc");
   ```
   This breaks for `cc`, `gcc` (no triple), versioned binaries like `riscv32-esp-elf-gcc-8.4.0`, or wrappers like `ccache gcc`. The `-print-sysroot` branch and the host-GCC path don't care about the binary name. For host GCC, `triple` would become e.g. `"x86_64-linux-gnu"` only if the binary is named that way; if it's just `gcc`, `triple` becomes `""` and `candidate` becomes `<prefix>`, which the `stdio.h` check will then probably reject. So it fails closed — but it's a sharp edge. Using the triple already in `CMakeConfigurer` (it has to know it) would be cleaner than parsing it back out of the filename.

3. **GCC localization.** `-print-search-dirs` output is generally not localized (it's documented machine-readable), but `String::from_utf8` will reject non-UTF-8 paths on Windows. Minor; same fragility exists in the primary branch.

4. **`compiler.file_name().unwrap_or_default()`** — if `derive_c_compiler()` ever returned a path with no filename (e.g. trailing slash), `name` becomes `""` and `triple` becomes `""`. Falls through harmlessly via the `stdio.h` check, but `expect()`-style on an internal invariant would read better than silently producing a bogus path.

5. **No caching.** Two extra `Command::new` invocations per build only on the slow path, so negligible — just noting it.

6. **The diff also changes `Command::new(compiler)` to `Command::new(&compiler)`** so `compiler` can be reused for the fallback. Necessary mechanical change, no issue.

## Verdict

Worth merging. The fix is well-scoped, the validation probe means it can't silently return garbage, and the targeted toolchains are real and commonly used by ESP developers. I'd ask for two small things before stamping: a one-line comment in the fallback noting the expected `<prefix>/lib/gcc/<triple>/<version>` layout it's reversing (so the four `.parent()` calls aren't a magic number for the next reader), and ideally sourcing the target triple from the configurer state rather than parsing the compiler binary's filename.

andrico21:
Pushed an update addressing both points.

1. Added a comment above the 4-parent walk noting the expected `<prefix>/lib/gcc/<triple>/<version>` layout.
2. Replaced the `compiler.file_name()` parsing with a new `derive_gcc_target_triple()` helper on `CMakeConfigurer` that mirrors the ESP cross-toolchain arms in `derive_forced_c_compiler`. The fallback now exits cleanly via `None` for host GCC or unforced RISC-V instead of producing a bogus path. The `stdio.h` probe is preserved as a safety net.

Verified locally with a downstream esp32c6 cross-build (`force_esp_riscv_gcc=true`, PIO `toolchain-riscv32-esp` on PATH) — links cleanly.

The non-symmetry between primary and fallback (primary may return `<prefix>/<triple>/sysroot`, fallback returns `<prefix>/<triple>`) is intentional and matches the layout of the targeted PIO toolchains. Happy to address if you'd prefer otherwise.

Issue #102, opened 2026-01-19
Hardware Acceleration tracking issue


AnthonyGrondin:
This issue keeps track of the current status of Hardware Acceleration support for ESP32.

**Legend:** If a cell contains an **em dash (—)** this means that the particular peripheral is not present for a chip. A **check mark (✓)** means that some hardware acceleration implementation exists. An **Aegean check mark (𐄂)** means that the chip has the crypto peripheral but implementation is currently impossible for `esp-mbedtls`.

| Peripheral | ESP32 | ESP32-C2 | ESP32-C3 | ESP32-C6 | ESP32-H2 | ESP32-S2 | ESP32-S3 | 
| :----------: | :--------: | :-----: | :------: | :------: | :------: | :------: | :------: | 
| AES |   | &mdash; |   |   |   |   |   |
| ECC | &mdash; |  | &mdash; |  |  | &mdash; | &mdash; |
| MD5* |   |   |   |   |   |   |   |
| RSA (Modular multiplication) |   | &mdash; |   |   |   |   |   |
| RSA (Modular exponentiation) | ✓ (512 bits..) | &mdash; | ✓ (3072 bits) | ✓ (3072 bits) | ✓ (3072 bits) | ✓ (4096 bits) | ✓ (4096 bits) |
| SHA1   | 𐄂 | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ |
| SHA224 | &mdash; | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ |
| SHA256 | 𐄂 | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ |
| SHA384 | 𐄂 | &mdash; | &mdash; | &mdash; | &mdash; | ✓ | ✓ |
| SHA512 | 𐄂 | &mdash; | &mdash; | &mdash; | &mdash; | ✓ | ✓ |

Note*: MD5 is a ROM-based software implementation. Verify whether it provides meaningful performance benefits and code-size savings compared to alternatives.

Implementation PRs:

 - https://github.com/esp-rs/esp-mbedtls/pull/20
 - https://github.com/esp-rs/esp-mbedtls/pull/24
 - https://github.com/esp-rs/esp-mbedtls/pull/46

ivmarkov:
I'm now really confused. 

What do you mean by "ECC"? I thought you meant **the ECDSA** peripheral? Is there more to elliptic curves w.r.t. hardware acceleration?

What I said is:
[...]

PR #129, opened 2026-05-30
feat: add ESP32-C5 hardware acceleration support

andrico21:
## Summary

Add an `esp32c5` feature to both `mbedtls-rs` and `mbedtls-rs-sys`, enabling ESP32-C5 hardware-accelerated mbedTLS.

ESP32-C5 shares the `riscv32imac-unknown-none-elf` target triple with C6, so it reuses the existing prebuilt static libraries — no new artifact build is required. The crypto peripherals (SHA, RSA, AES, ECC) are functionally compatible with the existing ESP hook backends.

## Change

1. **Cargo feature wiring** (mirrors the existing C3/C6/H2 pattern):
   - `mbedtls-rs-sys/Cargo.toml`: `esp32c5 = ["esp-hal/esp32c5", "crypto-bigint", "time"]`
   - `mbedtls-rs/Cargo.toml`: `esp32c5 = ["mbedtls-rs-sys/esp32c5"]`

2. **Op4096 cfg gate** (`mbedtls-rs-sys/src/hook/backend/esp/exp_mod.rs`): add `esp32c5` to the three `cfg(not(any(...c3, ...c6, ...h2)))` / one `cfg(any(...))` guards around the U4096 path. ESP32-C5's RSA peripheral, like C3/C6/H2, tops out at 3072-bit operands — `esp-hal::rsa::operand_sizes::Op4096` is not generated for the C5 chip, so without this gate `cargo build --features esp32c5` fails with:

   ```
   error[E0425]: cannot find type `Op4096` in module `operand_sizes`
      --> mbedtls-rs-sys/src/hook/backend/esp/exp_mod.rs:264:32
   ```

   `SOC_RSA_MAX_BIT_LEN = 3072` is correctly applied to C5 by the second cfg toggle.

## Validation

Tested against this branch rebased on current `main`, building the downstream `examples-esp`-style crate for all three RISC-V flagship chips:

| Chip | Target triple | Build | Notes |
|---|---|---|---|
| esp32c3 | riscv32imc-unknown-none-elf | ✅ | unchanged |
| esp32c6 | riscv32imac-unknown-none-elf | ✅ | unchanged |
| esp32c5 | riscv32imac-unknown-none-elf | ✅ | new |

Prebuilt static libs for `riscv32imac-unknown-none-elf` are reused for C5 — no new artifact in `mbedtls-rs-sys/lib/` needed.

## Notes / Follow-ups

- `examples/esp/Cargo.toml` does not currently include an `esp32c5` feature (only `esp32c2`/`c3`/`c6`). Happy to add that in a follow-up if desired; left out here to keep this PR strictly scoped to the crate-level feature add.
- Related issue: #102 (multi-chip support tracker).

## Scope

Three-file change. No public API breakage. No effect on existing chip targets.


ivmarkov (changes requested):
@andrico21 Yes, fix seems reasonable, but:
- Please add the esp32c5 build to matrix, or else CI can't test if your changes build correctly https://github.com/esp-rs/mbedtls-rs/blob/df426fba764c54b9b90281573b7d2971491727ac/.github/workflows/ci.yml#L194
- As you yourself mention: examples are not having this feature yet either, please add it there as well

Basically, we really have to support it everywhere, end-to-end like all other esp32* features.

andrico21:
Pushed an update.

1. Added `esp32c5` to the Build-MCU CI matrix — this validates the actual code changes end-to-end.

2. For `examples/esp/Cargo.toml`: the `[patch.crates-io]` block pins esp-hal at `rev=8a9fd5e`, which ships esp-radio 0.17. That release doesn't expose an `esp32c5` feature (added in 0.18, April 2026). Adding an `esp32c5 = [...]` line that references `esp-radio/esp32c5` would break feature resolution at manifest-load time.

I left the feature in place as a commented-out line with a note explaining the dep gap so it's trivially re-enabled once the patch rev is bumped. Happy to bump the rev in this PR if you'd prefer that path — it's a one-line change but touches all the other example chips too, so I figured it deserved a separate PR / discussion.

ivmarkov:
> 2. For `examples/esp/Cargo.toml`: the `[patch.crates-io]` block pins esp-hal at `rev=8a9fd5e`, which ships esp-radio 0.17. That release doesn't expose an `esp32c5` feature (added in 0.18, April 2026). Adding an `esp32c5 = [...]` line that references `esp-radio/esp32c5` would break feature resolution at manifest-load time.

Hold on then. I'll merge a small subset of #122 which **only** updates this crate to esp-hal 1.1 which will unlock the full c5 support everywhere. I'll ping you shortly once I'm done.

ivmarkov:
https://github.com/esp-rs/mbedtls-rs/pull/130 - If build succeeds, I'll merge and then this should unlock the changes in this PR.

ivmarkov:
@andrico21 OK #130 is now in, so I guess you can complete this PR.

andrico21:
Rebased onto current main (now includes #128, #130, #131).

Since #130 dropped [patch.crates-io] from `examples/esp` and bumped esp-radio to 0.18 — which exposes the `esp32c5` feature natively — uncommented the `esp32c5` entry in the example. CI should now build the example for C5 end-to-end.

No other changes.

ivmarkov:
esp32c5 build fails though?

andrico21:
Yes — root cause is an upstream esp-hal gap, not the PR's hw-accel changes.

The library itself (root `cargo build --features esp32c5`) passes. The failure is in `Clippy - Examples` because `examples/esp` enables the `hook-wall-clock` feature, which compiles `mbedtls-rs-sys/src/hook/backend/esp/wall_clock.rs`. That file calls `esp_hal::rtc_cntl::Rtc::current_time_us()`, gated upstream by `#[cfg(lp_timer_driver_supported)]`. esp32c5 has a `rtc/esp32c5.rs` chip module but the LP_TIMER driver isn't wired yet (no `lp_timer_driver_supported` cfg for c5 in esp-hal 1.1).

Additionally `examples/esp/src/bin/time_verification.rs` directly calls `rtc.current_time_us()` and `rtc.set_current_time_us()`, so even gating the wall_clock module wouldn't fix the example.

Options:
1. Scope this PR down to library-only c5 support: drop the matrix entry + example feature line, keep `mbedtls-rs[-sys]` features and `exp_mod.rs` cfg gates. End-to-end example CI lands once esp-hal adds c5 LP_TIMER.
2. Wait on this PR until esp-hal c5 LP_TIMER lands, then keep current scope.
3. Some workaround I'm missing (you know esp-hal c5 status better than I do).

What's your preference?

ivmarkov:
Isn't it possible to gate the `wall_clock.rs` implementation **and** the `time_verification.rs` example so that the example is not built for the c5 (or is build for the c5 but panics at runtime with a message "esp32c5 is not yet supported" - or similar)?

andrico21:
> Isn't it possible to gate the `wall_clock.rs` implementation **and** the `time_verification.rs` example?

check, updated. Thanks for quick replies btw :)

andrico21:
Pushed the gating. Three sites, all cfg-scoped to `feature = "esp32c5"`:

- `mbedtls-rs-sys/src/hook/backend/esp.rs`: `wall_clock` module excluded for c5.
- `examples/esp/src/bootstrap.rs`: RTC time-set + `EspRtcWallClock` hook skipped for c5. Tuple signature unchanged (`Rtc::new` works), so the other 7 examples build with no edits.
- `examples/esp/src/bin/time_verification.rs`: panics at runtime on c5 when the test reaches the RTC time-manipulation step.

Comment on each site points back at the underlying esp-hal gap (`lp_timer_driver_supported` not set for c5 in v1.1) so this is easy to revert once upstream lands the c5 LP_TIMER driver.

Local `cargo clippy --release --no-default-features --features esp32c5 -p mbedtls-rs-sys --target riscv32imac-unknown-none-elf` clean; examples crate clippy needs cmake which I don't have set up here, so CI will be the source of truth on the C build path.

andrico21:
CI surfaced a second pre-existing c5 incompatibility in `bootstrap.rs`: `esp_hal::rng::Trng` / `TrngSource` are also missing on c5 (gated by `rng_trng_supported` cfg, which esp-hal v1.1 doesn't set for c5). Same root cause as the RTC issue — chip drivers not yet wired.

Updated approach (force-pushed `e5ab61c`): cfg-gate the entire body of `bootstrap_stack` for c5 with a runtime-panic stub, and gate every c5-incompatible import + the `connection`/`net_task`/`wait_ip` helpers. This is more robust than chasing individual peripherals — when esp-hal lands more c5 drivers, the diff to revert is one cfg-block.

c5 bins now compile and link; they panic immediately on boot with a clear message. All non-c5 chips unchanged.

ivmarkov:
I'll merge but please note that without a TRNG on the esp32c5 the lib of course can be used, but won't be very secure.

andrico21:
Thanks for the merge! Noted on the TRNG/security caveat — our example bootstrap panics on c5 today, so users won't hit a silently-insecure TLS session via the examples. Once esp-hal wires the c5 TRNG we'll drop the cfg-gate.

PR #135, opened 2026-06-01
Add stdint.h as it is not always available

ivmarkov:
With certain clang installations (the one from Pigweed which is enabled during the rs-matter run against the YAML and Python integration tests from the C++ SDK) we don't really have the `stdint.h` header either.

Therefore, add that one as well to our custom sysroot.


ivmarkov:
Will merge directly because this should've been part of #119.

Also I just tested it as part of https://github.com/project-chip/rs-matter/pull/471

bjoernQ (approved):

PR #137, opened 2026-06-02
Turn the cargo xtask gen tgt into a glorified cargo wrapper

ivmarkov:
This is done for symmetry with OpenThread (https://github.com/esp-rs/openthread/pull/70), which has an identical build system. Although for `mbedtls-rs-sys` this change is not strictly necessary, as it does not depend on the C include directories of any other Rust crate.

But doing this means we won't surprise ourselves in 1y that the otherwise identical on the surface `cargo xtask gen` tasks in the two crates work in a completely different way.

As part of this PR, the support for using GCC was also slightly extended to support the Ubuntu/Debian system GCC, which has a slightly different layout than the typical crosstools-NG one.

=====

I've also added the thumbv6 (RPI Pico and others) and thumbv7 (nrf52 and others) pre-built targets because why not.

For now, neither supports any hardware acceleration
- For the Pico that would remain like that as it does not have any crypto hardware.
- For the NRF52 (and later) they do have crypto hardware, but I'm not sure to what extent it is documented...

Also, would be good to have examples for those too, which is tracked separately as #138


ivmarkov:
(Updated)

I also had to move the `mbedtls_platform_zeroize` symbol definition (that symbol, being a "platform" one is left as an exercise to the reader by MbedTLS) to a macro inside `mbedtls-rs-sys`, because otherwise if both `openthread` and `rs-matter` try to define it (as it happens now), we have a conflict.

Similarly to `calloc`, `free` etc, this symbol should always be defined by the final binary crate of the user, and never by "in-between" library crates like `mbedtls-rs` (the non-sys TLS/SSL IO crate), `openthread` and `rs-matter`.

The new `define_zeroize!` macro allows a one-line definition of the symbol by the user.

ivmarkov:
Changing temporarily to Draft as I am ATM chasing a huge size regression caused either by this PR, or the commit before it. Will un-draft once the root cause is clear.

ivmarkov:
> Changing temporarily to Draft as I am ATM chasing a huge size regression caused either by this PR, or the commit before it. Will un-draft once the root cause is clear.

@bjoernQ This is ready for review. (And there are ~3 to 4 more PRs ready on top of this coming.)

The size regression is first of all not caused by this PR, but by https://github.com/esp-rs/openthread/pull/70 in openthread, and is furthermore addressed by 
- https://github.com/esp-rs/mbedtls-rs/pull/139
- and https://github.com/esp-rs/openthread/pull/73

Proof:
- https://github.com/sysgrok/rs-matter-embassy/pull/55#issuecomment-4630498652

... which shows that we even have some small gains w.r.t. flash size once both `rs-matter` and `openthread` are configured to use `mbedtls-rs-sys` rather than `openthread` using its own MbedTLS private copy, while `rs-matter` using the RustCrypto backend, as it used to be before.

bjoernQ (approved):

PR #139, opened 2026-06-05
Fine-grained control over what MbedTLS algorithms get compiled-in

ivmarkov:
The idea is to have a bunch of features - one feature per algorithm so to say - which control which algorithms get compiled-in. These features roughly correspond 1:1 to the existing MbedTLC C `#define` toggles.

Why is this necessary? Same reason why the MbedTLC C `#define` toggles exist in the first place:
- Linker GC dead code removal **cannot** do this automatically, because a lot of the MbedTLC C algorithms are actually registered in global tables, and are thus referenced even when such an algorithm might not actually be used.

NOTE: Should merge **AFTER** #137


ivmarkov:
@bjoernQ This is now also ready for review, but please do not merge until the CI had auto-rebuilt and pushed new version of the `.a` libs (this is part of the CI, for PRs with the `rebuild-libs` label).

mbedtls-rs-sys/Cargo.toml
```diff
@@ -47,6 +60,169 @@ esp32h2 = ["esp-hal/esp32h2", "crypto-bigint", "time"]
...
+kex-dhe-psk = []
+kex-ecdhe-psk = []
+kex-ecjpake = []
+
+# ---------------------------------------------------------------------------
+# Platform alternatives
+# ---------------------------------------------------------------------------
+# Replace MbedTLS's built-in `mbedtls_platform_zeroize` with a user-supplied
+# one (define `MBEDTLS_PLATFORM_ZEROIZE_ALT` and provide the symbol via
+# `define_zeroize!`). OFF by default: MbedTLS's own zeroize is portable and
```
bjoernQ: isn't this macro removed by this PR?

bjoernQ (approved):
LGTM

mbedtls-rs-sys/Cargo.toml
ivmarkov: True! Addressed by https://github.com/esp-rs/mbedtls-rs/pull/139/changes/fa458c8d9778b45c058ce499e6c1de9b1bbfeb3a

PR #142, opened 2026-06-12
Expose granular algorithm selection; stop forcing the full `tls` profile

andrico21:
Today `mbedtls-rs` hard-wires `mbedtls-rs-sys`'s full `tls` profile, so every consumer compiles in the entire algorithm surface (all curves, RSA, DHM, every cipher mode) even when they only ever speak one suite. There's no way to trim from the high-level crate, since Cargo features are additive: a downstream crate can't subtract the `tls` that `mbedtls-rs` turns on.

That hurts on memory-constrained bare-metal targets. Concrete case: an ESP32-C5 firmware with 384 KB SRAM, where the TLS 1.3 handshake already needs ~46 KB of contiguous heap. The unused EC curve and ciphersuite tables are ~8 KB of pure `.data` we can't spare, and the deployment is P-256 / ECDSA only; RSA, DHM, brainpool and the secp-k1 / 521 curves are all dead weight.

This PR makes the algorithm surface selectable from `mbedtls-rs`, mirroring what `mbedtls-rs-sys` already exposes (#139):

- **`profile-tls` is a default feature** (= today's full surface), so existing default consumers are unchanged.
- The `tls` / `openthread` / `matter` bundles and every individual leaf (`alg-*`, `curve-*`, `cipher-mode-*`, `kex-*`, `pk`, `drbg-*`, `pem-*`, `x509-*`, ...) are re-exposed as thin 1:1 passthroughs, with no new logic.
- The hard-wired `features = ["tls"]` on the sys dependency is dropped; the profile now flows through these features.

Measured result downstream, with a P-256-only profile and identical numbers across ESP32-C3/C5/C6: **-7.8 KB `.data`**. On the C5 that converts straight into ~**+8 KB of usable task stack** (the linker hands freed DRAM to the stack region). HTTPS and the SHA-256 OTA FFI path build clean; RSA is fully absent from the image.

**Behavior note:** consumers using `default-features = false` previously still received the full `tls` set (it was forced regardless of their feature selection). They'll now need to opt into a profile or specific leaves. That's the intended fix, but worth flagging as a breaking change for that subset.

Happy to add matching feature docs to the README if you'd like them in the same PR.

### DIFF
--- a/mbedtls-rs/Cargo.toml
+++ b/mbedtls-rs/Cargo.toml
@@ -12,9 +12,98 @@ readme = "README.md"
 harness = false
 
 [features]
-default = ["edge-nal"]
+# `profile-tls` reproduces the historical full algorithm surface and is ON by
+# default, so existing consumers are unchanged. Consumers wanting a smaller build
+# set `default-features = false` and select a curated bundle (`tls`/`openthread`/
+# `matter`) or any individual `alg-*`/`curve-*`/`cipher-mode-*`/`kex-*`/… leaf below.
+default = ["edge-nal", "profile-tls"]
 defmt = ["dep:defmt", "mbedtls-rs-sys/defmt"]
 log = ["dep:log", "mbedtls-rs-sys/log"]
+
+# ─── Algorithm profile selection (granular passthrough to mbedtls-rs-sys) ───
+profile-tls = ["mbedtls-rs-sys/tls"]
+
+# Curated bundles
+openthread = ["mbedtls-rs-sys/openthread"]
+matter = ["mbedtls-rs-sys/matter"]
+tls = ["mbedtls-rs-sys/tls"]
+
+# Individual algorithm leaves (one per independently-linkable MbedTLS unit)
+alg-sha256 = ["mbedtls-rs-sys/alg-sha256"]
+alg-sha512 = ["mbedtls-rs-sys/alg-sha512"]
+alg-sha1 = ["mbedtls-rs-sys/alg-sha1"]
+alg-md5 = ["mbedtls-rs-sys/alg-md5"]
+alg-ripemd160 = ["mbedtls-rs-sys/alg-ripemd160"]
+alg-aes = ["mbedtls-rs-sys/alg-aes"]
+alg-des = ["mbedtls-rs-sys/alg-des"]
+alg-aria = ["mbedtls-rs-sys/alg-aria"]
+alg-camellia = ["mbedtls-rs-sys/alg-camellia"]
+alg-chacha20 = ["mbedtls-rs-sys/alg-chacha20"]
+cipher-mode-cbc = ["mbedtls-rs-sys/cipher-mode-cbc"]
+cipher-mode-cfb = ["mbedtls-rs-sys/cipher-mode-cfb"]
+cipher-mode-ofb = ["mbedtls-rs-sys/cipher-mode-ofb"]
+cipher-mode-ctr = ["mbedtls-rs-sys/cipher-mode-ctr"]
+cipher-mode-xts = ["mbedtls-rs-sys/cipher-mode-xts"]
+alg-gcm = ["mbedtls-rs-sys/alg-gcm"]
+alg-ccm = ["mbedtls-rs-sys/alg-ccm"]
+alg-cmac = ["mbedtls-rs-sys/alg-cmac"]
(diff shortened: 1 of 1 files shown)


andrico21:
Update: this profile is now runtime-validated on hardware.

I built an ESP32-C5 firmware that consumes the EC-only profile through these features (TLS 1.3, P-256 + x25519; RSA/DHM and unused curves dropped) and exercised the full stack:

- HTTPS server: browser plus `curl -vk`, with both the embedded P-256 cert and an NVS-uploaded cert/key.
- MQTTS client with strict verification (CA loaded, `AuthMode::Required`) against an EC broker chain.
- OTA firmware update over HTTPS, which also exercises the raw `mbedtls_sha256_*` FFI path.

All three work. Footprint vs the full `tls` profile, identical across C3/C5/C6: -7,792 B `.data` (the curve/ciphersuite tables), which on the C5 frees ~8 KB of task stack. The `mbedtls_sha256_*` symbols are retained and `mbedtls_rsa_*` is fully absent from the image.

mbedtls-rs/Cargo.toml
```diff
@@ -12,10 +12,99 @@ readme = "README.md"
...
+# `profile-tls` reproduces the historical full algorithm surface and is ON by
+# default, so existing consumers are unchanged. Consumers wanting a smaller build
+# set `default-features = false` and select a curated bundle (`tls`/`openthread`/
+# `matter`) or any individual `alg-*`/`curve-*`/`cipher-mode-*`/`kex-*`/… leaf below.
+default = ["edge-nal", "profile-tls"]
 defmt = ["dep:defmt", "mbedtls-rs-sys/defmt"]
 log = ["dep:log", "mbedtls-rs-sys/log"]
 
+# ─── Algorithm profile selection (granular passthrough to mbedtls-rs-sys) ───
+profile-tls = ["mbedtls-rs-sys/tls"]
```
ivmarkov: What is the difference between "profile-tls" and "tls" below. Remove.

mbedtls-rs/Cargo.toml
```diff
@@ -12,10 +12,99 @@ readme = "README.md"
 harness = false
 
 [features]
-default = ["edge-nal"]
+# `profile-tls` reproduces the historical full algorithm surface and is ON by
+# default, so existing consumers are unchanged. Consumers wanting a smaller build
+# set `default-features = false` and select a curated bundle (`tls`/`openthread`/
+# `matter`) or any individual `alg-*`/`curve-*`/`cipher-mode-*`/`kex-*`/… leaf below.
+default = ["edge-nal", "profile-tls"]
```
ivmarkov: I wonder if I try to build the mbedtls-rs crate after these changes, with `cargo build --no-default-features` whether it would build or break with a build error.

If it breaks that would not be nice. Sure, you could make the same argument for `mbedtls-rs-sys`, but the sys one is at least a very low level, unlike this one.

Where I'm going with that is that perhaps we need a minimal, non-negotiable set of features enabled on `mbedtls-rs-sys` **from** `mbedtls-rs` so that `mbedtls-rs` at least compiles.

andrico21:
Both addressed:

1. Removed the redundant `profile-tls`; `default` now uses `tls` directly.

2. Good catch on `--no-default-features`. I added a small unconditional floor of `mbedtls-rs-sys` features on the dependency so `mbedtls-rs` still compiles with no bundle selected: the parts its own Rust API uses (SSL engine, X.509/PK/PEM parse, RNG) plus a coherent minimal EC TLS 1.2/1.3 stack (ECDHE-ECDSA over P-256, AES-GCM, SHA-256, HKDF). Consumer-selected bundles and leaves still union on top. I kept this to SHA-256 only, since SHA-512 is not needed for the minimum.

If you'd rather the no-default build mean strictly "the Rust crate compiles" rather than "minimally usable TLS", the next thing to test in CI would be dropping TLS 1.2 from the floor. Happy to go either way.

ivmarkov:
> If you'd rather the no-default build mean strictly "the Rust crate compiles" rather than "minimally usable TLS", the next thing to test in CI would be dropping TLS 1.2 from the floor. Happy to go either way.

No "minimally usable TLS" is even better and I doubt anybody would decide to use mbedtls without the nist256p1 curve any time soon (or without sha256), so LGTM.

PR #144, opened 2026-06-13
Allow tuning the TLS record buffer sizes to reduce per-connection RAM

andrico21:
## Problem

MbedTLS allocates `MBEDTLS_SSL_IN_CONTENT_LEN` + `MBEDTLS_SSL_OUT_CONTENT_LEN` record-plaintext buffers **per TLS connection**. Both default to `16384`, so every connection costs ~32 KB of heap — the dominant RAM cost on memory-constrained MCUs. Negotiating a smaller Max-Fragment-Length does **not** shrink these buffers (they are `calloc`d at the compiled `*_CONTENT_LEN` size in `mbedtls_ssl_setup`); only lowering `*_CONTENT_LEN` does. And since the committed prebuilt `.a` libraries bake in `16384`, a header/user-config override alone cannot resize them.

## Change

`ssl-in-content-len-<N>` / `ssl-out-content-len-<N>` cargo features (`N` one of `256, 512, 1024, 2048, 4096, 8192, 16384`), re-exposed 1:1 by `mbedtls-rs`:

```toml
mbedtls-rs = { version = "...", features = ["ssl-in-content-len-4096", "ssl-out-content-len-2048"] }
```

- **Additive — largest wins:** Cargo features unify across the dependency graph, so if several `ssl-in-content-len-*` (or `-out-`) are enabled, the largest is used — always safe for the consumer that asked for a smaller buffer.
- Applied in `features::apply_features`, so the selection feeds **both** the prebuilt-validity check and the actual compile config. It is absent from the prebuilt reference (`prebuilt_features_config`), so a non-default size registers as a config delta and **forces an on-the-fly MbedTLS rebuild** (needs `clang`/`cmake`/`ninja`).
- **Default-preserving:** selecting nothing (or `-16384`) keeps MbedTLS's default and reuses the committed prebuilt libraries — existing consumers are unaffected.
- Skipped under the `prebuilt` profile, so `xtask` always regenerates default-sized artifacts (and the selection never trips `build.rs`'s desync guard).

Docs added to the `mbedtls-rs-sys` README, including the `IN`-too-low / `OUT`-too-low handshake-failure caveats.

> Reworked per @ivmarkov's review — cargo features instead of env vars, biggest-prevails — following the `openthread-sys` `heap-int-*` convention.

## Verification

Built on host, target `riscv32imc-unknown-none-elf`:

<details>
<summary><b>Build evidence</b> — click to expand</summary>

```text
# default (no feature) -> prebuilt reused, no rebuild
$ cargo build -p mbedtls-rs-sys --target riscv32imc-unknown-none-elf            -> Finished (reused)

# features -> prebuilt rejected + on-the-fly rebuild + #defines emitted
$ cargo build -p mbedtls-rs-sys --features ssl-in-content-len-4096,ssl-out-content-len-2048 --target riscv32imc-unknown-none-elf
# generated mbedtls_rs_sys_user_config.h:
#define MBEDTLS_SSL_IN_CONTENT_LEN 4096
#define MBEDTLS_SSL_OUT_CONTENT_LEN 2048

# biggest prevails: ssl-in-content-len-2048 + ssl-in-content-len-4096 -> IN=4096
#define MBEDTLS_SSL_IN_CONTENT_LEN 4096

# ssl-in-content-len-16384 (= default) -> reused, no #define
```

</details>

### Verified on-target (esp32c3 in QEMU)

A no-network firmware measured the heap allocated per `Session::new` on a real esp32c3 boot — `Session::new` runs `mbedtls_ssl_setup` (which `calloc`s the IN/OUT record buffers); the handshake, which needs a peer, is never started:

| Config | heap allocated per `Session::new` |
|---|---|
| default (IN=16384, OUT=16384) | **43,276 B** |
| `ssl-in-content-len-4096` + `ssl-out-content-len-2048` | **16,652 B** |

That is **26,624 B less per session — exactly `(16384 − 4096) + (16384 − 2048)`**: the on-target heap delta drops by precisely the configured buffer reduction.

### DIFF
--- a/mbedtls-rs-sys/gen/features.rs
+++ b/mbedtls-rs-sys/gen/features.rs
@@ -540,6 +540,52 @@ pub fn apply_features(config: &mut MbedtlsUserConfig) {
     build_config(config, |feature| {
         std::env::var_os(format!("CARGO_FEATURE_{feature}")).is_some()
     });
+    apply_ssl_content_len_overrides(config);
+}
+
+/// The selectable `MBEDTLS_SSL_IN_CONTENT_LEN` / `MBEDTLS_SSL_OUT_CONTENT_LEN`
+/// sizes, each exposed as an `ssl-{in,out}-content-len-<N>` cargo feature (see
+/// Cargo.toml). Cargo features are additive, so when several are enabled across
+/// the dependency graph the LARGEST is used — a bigger record buffer is always
+/// safe for the consumer that asked for a smaller one. Keep in sync with the
+/// `ssl-*-content-len-*` feature lists in Cargo.toml.
+pub const SSL_CONTENT_LEN_SIZES: &[usize] = &[256, 512, 1024, 2048, 4096, 8192, 16384];
+
+/// Apply the consumer-selected TLS record-buffer sizes (the
+/// `ssl-{in,out}-content-len-<N>` features) to `config`.
+///
+/// These cap the per-connection record-plaintext buffers MbedTLS `calloc`s in
+/// `mbedtls_ssl_setup`. The largest enabled size per direction wins (additive
+/// features). No size selected — or `16384`, MbedTLS's default — leaves the
+/// macro undefined so the committed prebuilt libraries are reused; any smaller
+/// value is `#define`d, which (being absent from the prebuilt reference config
+/// in `prebuilt_features_config`) registers as a config delta and forces an
+/// on-the-fly MbedTLS rebuild. Applied only here, never to the prebuilt
+/// reference, so the override is always visible as a delta against the `.a`.
+fn apply_ssl_content_len_overrides(config: &mut MbedtlsUserConfig) {
+    // The committed prebuilt artifacts are always built at MbedTLS's default
+    // buffer size, so ignore the size features while generating them (xtask's
+    // `prebuilt` profile) — otherwise the desync guard in build.rs would reject
+    // its own output.
+    if std::env::var_os("CARGO_FEATURE_PREBUILT").is_some() {
+        return;
+    }
+
+    const DEFAULT_CONTENT_LEN: usize = 16384;
+
+    for ident in ["SSL_IN_CONTENT_LEN", "SSL_OUT_CONTENT_LEN"] {
+        let selected = SSL_CONTENT_LEN_SIZES
+            .iter()
+            .copied()
+            .filter(|n| std::env::var_os(format!("CARGO_FEATURE_{ident}_{n}")).is_some())
+            .max();
+
+        if let Some(value) = selected {
+            if value != DEFAULT_CONTENT_LEN {
+                config.set(ident, value.to_string());
+            }
+        }
+    }
 }
 
 /// Build the algorithm-selection config the *prebuilt* artifacts correspond to
--- a/mbedtls-rs-sys/Cargo.toml
+++ b/mbedtls-rs-sys/Cargo.toml
@@ -28,6 +28,29 @@ default = ["tls"]
 # hooks match this profile exactly.
 prebuilt = ["tls"]
 
+# ─── TLS record buffer sizing (per-connection RAM) ───
+# Cap MBEDTLS_SSL_IN/OUT_CONTENT_LEN below the 16384 default to shrink the
+# per-connection record buffers. Cargo features are additive, so when several
+# sizes are selected across the dependency graph the LARGEST wins (a bigger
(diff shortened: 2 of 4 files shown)


mbedtls-rs-sys/gen/features.rs
```diff
@@ -540,6 +540,58 @@ pub fn apply_features(config: &mut MbedtlsUserConfig) {
     build_config(config, |feature| {
         std::env::var_os(format!("CARGO_FEATURE_{feature}")).is_some()
     });
+    apply_ssl_content_len_overrides(config);
```
ivmarkov, line 543: Instead of fiddling with environment variables, I strongly suggest using features as follows:
- `ssl-in-content-len-256`
- `ssl-in-content-len-512`
-  `ssl-in-content-len-1024`
-  `ssl-in-content-len-2048`
- ...
-  `ssl-in-content-len-16384`

... and ditto for `ssl-out-content-len-XXX`

If the user defines multiple ssl-in-content-len or ssl-out-content-len, the biggest one should prevail.

See also [this](https://github.com/esp-rs/openthread/blob/14153ec0837b49316dd8f0ad4fded4e4655b8219/openthread-sys/Cargo.toml#L82) for further inspiration.

andrico21: Done - switched to `ssl-in-content-len-<N>` / `ssl-out-content-len-<N>` cargo features (`256 … 16384`), with the **largest selected size winning** across the dependency graph, following the `openthread-sys` `heap-int-*` convention. `mbedtls-rs` re-exposes them 1:1.

Re-verified:
- default → prebuilt reused (no rebuild); a smaller size → prebuilt rejected + on-the-fly rebuild + the right `#define`s; biggest-prevails (`ssl-in-content-len-2048` + `ssl-in-content-len-4096` → `IN=4096`); `ssl-in-content-len-16384` → reused.
- On a real **esp32c3 in QEMU**, the per-`Session` heap drops **43,276 → 16,652 B** with `ssl-in-content-len-4096` + `ssl-out-content-len-2048` (exactly the buffer reduction).

Thanks for the steer!

ivmarkov:
Hmmm, however this all could only work if you really control both ends (TLS client _and_ server) so that you can guarantee that the TLS nego payloads are small enough to fix into the reduced buffers? If you could confirm?

andrico21:
> Hmmm, however this all could only work if you really control both ends (TLS client _and_ server) so that you can guarantee that the TLS nego payloads are small enough to fix into the reduced buffers? If you could confirm?

In my case the payload is tiny and strictly bound to P-256 certs: HTTPS-server to control a VAC device with proper auth and MQTTS client for controls/telemetry. OTA part is also handled properly over HTTPS - so it worth to have such controls, I had to make such edits locally.

So, P-256/EC certs keep the server's Certificate message small, and that's the inbound record that sizes IN. With an RSA chain you'd need a larger IN (or the server fragmenting).

For the general case it stays a safe opt-in either way:
- OUT (ssl-out-content-len-*) needs no assumption about the peer, it only bounds what this device sends (its own cert as a server, or ClientHello / client cert as a client).
- An inbound record that doesn't fit aborts the handshake with an error, never a silent truncation, and the 16384 default is untouched, so nothing changes for anyone who doesn't opt in.

One honest gap: MFL is compiled in but the crate doesn't expose it yet, so the inbound side currently relies on a known/small-cert server. Wiring up mbedtls_ssl_conf_max_frag_len could relax that as a follow-up.

andrico21:
Correction to my comment above: the MFL remark was wrong.

I said wiring up `mbedtls_ssl_conf_max_frag_len` could relax the inbound/small-cert reliance. It can't. Per the MbedTLS docs (`ssl.h`, on `mbedtls_ssl_conf_max_frag_len`):

> With TLS, this currently only affects ApplicationData (sent with `mbedtls_ssl_read()`), not handshake messages.

The server's `Certificate` is a handshake message, so MFL does **not** reduce the buffer needed to receive it, and does not relax the inbound side at all. It only caps ApplicationData record payload (and it's TLS 1.2 / RFC 6066 only; TLS 1.3 uses `record_size_limit`). So `ssl-in-content-len-*` remains the only lever that actually sizes the handshake-receive buffer.

I've since exposed MFL as its own knob in #145, documented honestly as an ApplicationData cap that is explicitly **not** a buffer-sizing mechanism. The rest of my earlier comment (P-256 keeping the Certificate small, `OUT` needing no peer assumption, a too-small inbound record aborting the handshake rather than truncating) still holds.

PR #146, opened 2026-06-13
Fix reachable panics in the ESP exp_mod hook

andrico21:
## What

Fixes a reachable abort cluster in the ESP hardware modular-exponentiation hook, `EspExpMod::exp_mod` (the backend for `mbedtls_mpi_exp_mod`, covering RSA and finite-field DH).

The hook panicked instead of returning an error on several inputs. The most serious is reachable from negotiated handshake parameters: a **3072-bit modulus (96 limbs)** passes the size guard but matches no hardware operand-size arm and hits `unreachable!()` -> abort. `ffdhe3072` and DHE-RSA are compiled into the default config, so a peer (or selected ciphersuite) can drive this. Other inputs aborted via `todo!()`, `panic!()`, `unwrap!()`, and `assert_eq!()`.

## How

The C-ABI shim `mbedtls_mpi_exp_mod` already maps a Rust `Err` to a negative MbedTLS code, so the fix makes the hook **return errors / fall back to software** rather than abort:

- **Routing extracted to a testable module.** A new `esp_hal`-free `backend::esp_exp_mod_route` decides, from the operand limb count, whether the RSA peripheral can serve a request (`Route::Hw(HwKind)`) or it must use software (`Route::Soft`). The hardware `match` now dispatches on `HwKind` and is exhaustive: there is no `unreachable!()`, and any operand size the peripheral cannot serve (including 1536/2560/3072/3584-bit) falls back to `mbedtls_mpi_exp_mod_soft`.
- **Operands validated before any limb dereference.** A null or non-positive/even modulus, a negative exponent, and null/internally-inconsistent MPIs now return `MBEDTLS_ERR_MPI_BAD_INPUT_DATA` instead of panicking.
- **Negative base routed to software** (which implements the sign semantics the old hardware compensation branch intended). This also removes an `assert_eq!(x.private_s, 1)` that panicked on a negative base and made the sign-compensation branch below it dead code.
- **`unwrap!(merr!(...))` replaced with `merr!(...)?`**, and temporary MPIs are wrapped in a small RAII guard so early returns free them.

Behavior for valid hardware operand sizes (256/512/1024/2048-bit, plus 384/4096-bit where the chip supports them) is unchanged.

## Testing

- **Host integration test** (`tests/exp_mod_route.rs`) exercises the production `hw_route` and asserts that supported sizes route to hardware and unsupported sizes - including the 3072-bit class behind this bug - route to software. The internal `_route-test` feature enables `crypto-bigint` without `esp-hal` so the pure routing logic builds and runs on the host. `cargo test -p mbedtls-rs-sys --features _route-test,tls --test exp_mod_route` -> 3 passed.
- **ESP target build:** `cargo +nightly build -p mbedtls-rs-sys --features esp32c3,tls --target riscv32imc-unknown-none-elf -Zbuild-std=core,alloc` builds clean (the riscv `Op512`/`Op1024`/`Op2048` arms).
- A static scan confirms no `todo!`/`panic!`/`unreachable!`/`unwrap!`/`assert!` remain in `esp/exp_mod.rs`; clippy is clean on the lib and the test target.

## Notes

- The Xtensa operand arms (`Op256`/`Op384`/`Op4096`, on `esp32`/`esp32s2`/`esp32s3`) are covered by the host routing test, and CI builds them on those targets; I could not build them locally (no `esp` Rust toolchain available), so the host test was my local proxy for that path.
- `prec_rr` reuse still relies on the existing MbedTLS contract that a non-null cached value is valid for the same modulus; this PR does not change that contract.
- This is one of a few fixes coming out of a wider review of the hooks and async TLS paths; I expect to open the next PR (the async SSL-context aliasing fix) in the next few days.

### DIFF
--- a/mbedtls-rs-sys/Cargo.toml
+++ b/mbedtls-rs-sys/Cargo.toml
@@ -13,6 +13,15 @@ exclude = ["*.pyc", "mbedtls/framework", "mbedtls/programs", "mbedtls/tests"]
 [lib]
 harness = false
 
+# Host-only routing test (F-01). Gated on the internal `_route-test` feature so
+# the default `cargo test` (no `crypto-bigint`) skips this target entirely.
+# `harness = true` is the integration-test default and set here only for clarity
+# (the `[lib] harness = false` above governs the lib target, not `tests/`).
+[[test]]
+name = "exp_mod_route"
+required-features = ["_route-test"]
+harness = true
+
 [features]
 # The default profile is the full config (`tls`), so behavior is unchanged for
 # existing consumers and the pre-generated bindings/libs apply as-is. Select a
@@ -82,6 +91,12 @@ esp32c6 = ["esp-hal/esp32c6", "crypto-bigint", "time"]
 esp32h2 = ["esp-hal/esp32h2", "crypto-bigint", "time"]
 esp32s2 = ["esp-hal/esp32s2", "crypto-bigint", "time"]
 esp32s3 = ["esp-hal/esp32s3", "crypto-bigint", "time"]
+
+# Internal: host-only testing of the ESP exp_mod routing logic. Pulls in
+# `crypto-bigint` (for the operand-size limb constants) WITHOUT `esp-hal`, so
+# `backend::esp_exp_mod_route` and its integration test compile on the host.
+# Not for downstream use.
+_route-test = ["crypto-bigint"]
 
 # ---------------------------------------------------------------------------
 # Features controlling what MbedTLS algorithms gets compiled in.
--- a/mbedtls-rs-sys/src/hook/backend.rs
+++ b/mbedtls-rs-sys/src/hook/backend.rs
@@ -2,5 +2,12 @@
 pub mod embassy;
 #[cfg(feature = "esp-hal")]
 pub mod esp;
+#[cfg(all(not(feature = "esp-hal"), feature = "_route-test"))]
+pub mod esp_exp_mod_route;
+#[cfg(all(
+    feature = "esp-hal",
+    not(any(feature = "esp32c2", feature = "nohook-exp-mod"))
+))]
+pub(crate) mod esp_exp_mod_route;
 #[cfg(feature = "std")]
 pub mod std;
--- a/mbedtls-rs-sys/src/hook/backend/esp_exp_mod_route.rs
+++ b/mbedtls-rs-sys/src/hook/backend/esp_exp_mod_route.rs
@@ -1,0 +1,88 @@
+//! Pure routing decision for ESP modular-exponentiation: pick the hardware
+//! operand size for a given limb count, or fall back to software.
+//!
+//! This is intentionally free of any `esp_hal` dependency so it can be unit
+//! tested on the host. The production ESP backend in
+//! [`super::esp::exp_mod`](super::esp::exp_mod) calls [`Route::route_for`] to
+//! decide whether a request can be served by the RSA peripheral or must use the
+//! MbedTLS software path. Keeping it here (rather than inline in the
+//! `esp_hal`-gated module) is what makes the routing logic testable without an
+//! ESP target.
+
+use crypto_bigint::U1024;
+use crypto_bigint::U2048;
+#[cfg(not(feature = "esp32"))]
+use crypto_bigint::U256;
+#[cfg(not(feature = "esp32"))]
+use crypto_bigint::U384;
+#[cfg(not(any(
+    feature = "esp32c3",
+    feature = "esp32c5",
+    feature = "esp32c6",
+    feature = "esp32h2"
+)))]
+use crypto_bigint::U4096;
+use crypto_bigint::U512;
+
+/// A hardware-supported RSA operand size, identified by its limb count.
+///
+/// Each variant corresponds 1:1 to an `esp_hal::rsa::operand_sizes::OpN` and is
+/// gated by exactly the same chip features as the hardware path, so a variant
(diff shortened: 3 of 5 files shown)


andrico21:
Sorry for the large one bud

mbedtls-rs-sys/src/hook/backend/esp_exp_mod_route.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/// Decide how to serve a modular exponentiation of `num_words` 32-bit limbs.
+///
+/// `num_words` is the post-`calculate_hw_words` limb count (already rounded up
+/// to the 512-bit block on `esp32`). Any size the peripheral cannot serve,
+/// including in-range-but-unsupported sizes such as 96 limbs (3072-bit) on a
+/// chip whose RSA arms stop at 2048, returns [`Route::Soft`] instead of
+/// panicking.
+pub fn hw_route(num_words: usize) -> Route {
```
ivmarkov: rename to `route_for` and move as a method (a constructor fn of sorts) in `Route` which returns `Self`.

mbedtls-rs-sys/tests/exp_mod_route.rs
```diff
@@ -0,0 +1,48 @@
+//! Host tests for the ESP modular-exponentiation routing decision (F-01).
```
ivmarkov, line 1: A bit of a question of a personal taste but generally I prefer to have the unit tests _inside_ the actual code module.

In that case that would be `exp_mod_route`, as a `mod tests { ... }` inline sub-module.

mbedtls-rs-sys/src/hook/backend.rs
```diff
@@ -2,5 +2,12 @@
 pub mod embassy;
 #[cfg(feature = "esp-hal")]
 pub mod esp;
+#[cfg(all(not(feature = "esp-hal"), feature = "_route-test"))]
+pub mod esp_exp_mod_route;
```
ivmarkov, line 6: Do we need this to be pub?

mbedtls-rs-sys/Cargo.toml
```diff
@@ -74,14 +74,20 @@ hook-wall-clock = ["hook-timer"] # 'hook-wall-clock' feature requires 'hook-time
...
-esp32h2 = ["esp-hal/esp32h2", "crypto-bigint", "time"]
-esp32s2 = ["esp-hal/esp32s2", "crypto-bigint", "time"]
-esp32s3 = ["esp-hal/esp32s3", "crypto-bigint", "time"]
+esp32c3 = ["esp-hal/esp32c3", "dep:crypto-bigint", "time"]
+esp32c5 = ["esp-hal/esp32c5", "dep:crypto-bigint", "time"]
+esp32c6 = ["esp-hal/esp32c6", "dep:crypto-bigint", "time"]
+esp32h2 = ["esp-hal/esp32h2", "dep:crypto-bigint", "time"]
+esp32s2 = ["esp-hal/esp32s2", "dep:crypto-bigint", "time"]
+esp32s3 = ["esp-hal/esp32s3", "dep:crypto-bigint", "time"]
+
```
ivmarkov, line 94: Kinda nit but...: 

All of the above changes are unnecessary, because cargo allows both `dep:foo` and just `foo`. Also valid for the new `_route-test` line below.

In general, I tend to just use it without the dep: and only introduce the dep: prefix when there are **both** a feature named e.g. `foo` and an optional dependency crate named `foo` (and then you need to be explicit what you are referring to).

Also introducing "dep:" just for crypto-bigint and not doing it for "time" is not symmetrical so I suggest you just strip the newly-introduced "dep:".

ivmarkov (commented):
Overall LGTM but with some suggestions for small cleanups / reorgs.

andrico21:
Thanks for the review, agreed for the most of catches:

- route_for as a method on Route - agreed, cleaner.
- dep:crypto-bigint - you're right, fine to strip it. (I only added it because cargo errored once I referenced crypto-bigint from the new _route-test feature - the rule is it's all-bare or all-dep:, you can't mix, and I'd mixed it. Going all-bare, symmetric with time.)
- Gating the test (required-features = ["_route-test"]) - got it, accepted for changing.
- "32-bit limbs" doc - correct, crypto_bigint::*::LIMBS is word-sized (u64 on host, u32 on the ESP target). I'll reword to word-sized limbs and note they're the 32-bit mbedtls_mpi_uint limbs on-target.

On the inline mod tests (and the related "does it need to be pub?") - I'd actually pushed it into a separate integration test on purpose, and I don't think it can move inline here: mbedtls-rs-sys is #![no_std] with [lib] harness = false, so an in-crate #[cfg(test)] mod tests doesn't build as a test target (cargo test on the lib fails with #[panic_handler] required / main not found). The integration test works because it compiles as its own std crate against the lib. That's also why the module is pub on the _route-test line only - an integration test is a separate crate, so it needs pub to import route_for/Route/HwKind; production stays pub(crate).

If you'd rather have inline tests, that's doable but it's a broader change to the crate's test setup (flipping to std under cfg(test), or re-enabling the harness) - happy to do it separately if you want it, but it felt out of scope for this fix. Let me know which you prefer.

ivmarkov:
> The integration test works because it compiles as its own std crate against the lib. That's also why the module is pub on the _route-test line only - an integration test is a separate crate, so it needs pub to import route_for/Route/HwKind; production stays pub(crate)

Fsir enough, I totally forgot about no_std and test-harness=false.

Fine with me then!

Just mark all my other comments as resolved when they are and assess the couple Copilot produced.

PR #148, opened 2026-06-14
Surface transport errors and stop the blocking read busy-spin on EOF

andrico21:
Another small soundness/robustness fix for the blocking session, independent of the others.

## What

Two related bugs in the blocking BIO callbacks:

1. **Busy-spin on EOF.** `bio_receive` mapped a stream `Ok(0)` to `MBEDTLS_ERR_SSL_WANT_READ`. Per the `embedded-io` `Read` contract, a non-empty-buffer `Ok(0)` is EOF, and MbedTLS always passes a non-empty buffer. So on a half-closed peer (transport FIN with no TLS close-notify) the read loop treated `WANT_READ` as "retry" and span forever, pinning the CPU.

2. **Swallowed I/O errors.** Both callbacks discarded the real stream error (`Err(_) => 0`), and a `0` return reads to MbedTLS as a clean close, so a genuine transport error surfaced to the caller as a clean EOF rather than an error.

The async path already handled EOF correctly (via `wait_readable` returning `Ok(false)` -> `BrokenPipe`), so this is blocking-only.

## How

A BIO callback can only return an `i32`, so the real error is captured on the session and surfaced by the caller-side loops:

- Added `last_io_err: Option<ErrorKind>` to the blocking `Session`.
- `bio_receive`: `Ok(0)` -> `MBEDTLS_ERR_SSL_CONN_EOF` (was `WANT_READ`); `Err(e)` -> store `e.kind()` + `CONN_EOF`.
- `bio_send`: `Err(e)` -> store `e.kind()` + `CONN_EOF`; a non-empty-buffer `Ok(0)` (which `embedded-io` `Write` is not allowed to return, it must error `WriteZero`) -> store `WriteZero` + `CONN_EOF` rather than retrying `WANT_WRITE` (which would be the write-side mirror of the same spin).
- `read`/`write`/handshake/`close` drain the captured kind first and return `SessionError::Io(kind)`. For `read`, an abrupt close with no close-notify shows up as MbedTLS returning `0` (it special-cases a BIO `CONN_EOF` into a `0` return, see `ssl_msg.c`), which is mapped to `Io(BrokenPipe)`. A clean TLS close-notify still returns `Ok(0)`.
- `call_mbedtls` clears `last_io_err` on entry so a kind can never leak across calls, and `read` short-circuits an empty buffer to `Ok(0)` so it is never misreported as an abrupt close.

`CONN_EOF` is used because it is present in all the per-target prebuilt bindings (the `net_sockets` `MBEDTLS_ERR_NET_*` codes are not compiled into the bare-metal build) and MbedTLS treats it as terminal in `ssl_read`/`ssl_write`.

The public API is unchanged, no new dependency, no `mbedtls-rs-sys` change.

## A note for you

For the abrupt-close-without-close-notify case I went with `ErrorKind::BrokenPipe` (it matches what the async path returns for the same situation). `ErrorKind::ConnectionReset` is arguably a closer fit semantically. I left a comment at the mapping site flagging this as your call, happy to switch it if you prefer.

## Testing

The real `bio_receive`/`bio_send` and the loop mapping are private and driving the public `read` needs the MbedTLS C library plus a live handshake, so (mirroring `tests/async_split_aliasing.rs`) `tests/blocking_eof.rs` models the callback and loop-arm logic in isolation over a mock `embedded-io` stream and locks: EOF -> `CONN_EOF` (not `WANT_READ`), a stream error is captured and surfaced, the read loop maps a bare `0` to `BrokenPipe` while a captured kind wins, `bio_send` `Ok(0)`/`Err` map to `WriteZero`/the real kind, the empty-buffer short-circuit, and the `close` drain.

Also verified: `cargo build`/`clippy`/`fmt --check` on `mbedtls-rs`, and the esp32c3 and esp32c2 example builds (the latter via the nightly `-Zbuild-std` invocation).

### DIFF
--- a/mbedtls-rs/src/session/asynch.rs
+++ b/mbedtls-rs/src/session/asynch.rs
@@ -594,12 +594,12 @@ where
             {
                 MBEDTLS_ERR_SSL_WANT_READ => {
                     if !self.wait_readable().await.map_err(SessionError::from_io)? {
-                        return Err(SessionError::Io(ErrorKind::BrokenPipe));
+                        return Err(SessionError::Io(ErrorKind::ConnectionReset));
                     }
                 }
                 MBEDTLS_ERR_SSL_WANT_WRITE => {
                     if !self.wait_writable().await.map_err(SessionError::from_io)? {
-                        return Err(SessionError::Io(ErrorKind::BrokenPipe));
+                        return Err(SessionError::Io(ErrorKind::ConnectionReset));
                     }
                 }
                 // See https://github.com/Mbed-TLS/mbedtls/issues/8749
@@ -629,7 +629,7 @@ where
             {
                 MBEDTLS_ERR_SSL_WANT_READ => {
                     if !self.wait_readable().await.map_err(SessionError::from_io)? {
-                        return Err(SessionError::Io(ErrorKind::BrokenPipe));
+                        return Err(SessionError::Io(ErrorKind::ConnectionReset));
                     }
                 }
                 // See https://github.com/Mbed-TLS/mbedtls/issues/8749
@@ -663,7 +663,7 @@ where
             {
                 MBEDTLS_ERR_SSL_WANT_WRITE => {
                     if !self.wait_writable().await.map_err(SessionError::from_io)? {
-                        return Err(SessionError::Io(ErrorKind::BrokenPipe));
+                        return Err(SessionError::Io(ErrorKind::ConnectionReset));
                     }
                 }
                 // See https://github.com/Mbed-TLS/mbedtls/issues/8749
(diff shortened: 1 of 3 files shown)


ivmarkov (commented):
LGTM, but if you could indeed replace the BrokenPipe error with ConnectionReset - as suggested. In both the blocking and async code. I agree ConnectionReset is a bit closer to what is happening.

andrico21:
> LGTM, but if you could indeed replace the BrokenPipe error with ConnectionReset - as suggested. In both the blocking and async code. I agree ConnectionReset is a bit closer to what is happening.

done

PR #152, opened 2026-06-15
Document cancel safety on the async TLS API

andrico21:
Hi Ivan, a small docs-only PR to bring the async API in line with `RUST_GUIDELINES.md` G§5 (every async fn should carry a cancel-safety annotation).

## What

The async `Session` API (and its split halves) had no cancel-safety annotations. As G§5 notes, cancel safety is not expressible in the type system, so a refactor that moves a previously-sequential call into a `select!` / `timeout` / `abort` arm can silently turn correct code into a partial-write / corrupt-stream bug. The most concrete case is `write`: MbedTLS `ssl_write` must be retried with the same arguments after `WANT_WRITE`, so dropping a `write` future mid-flight and re-issuing with a different buffer makes the peer receive the previously-buffered record while the new call reports success.

## How

Annotate every async fn in `session/asynch.rs` as `NOT cancel-safe` (with the two genuinely cancel-safe `NoRead`/`NoWrite` pass-throughs marked as such). Every public inherent async fn also gets a `# Cancel safety` rustdoc section, since the `//` comments G§5 uses are invisible on the rendered docs. The `write` / `read` / `flush` wording is deliberately precise:

- `write`: after `WANT_WRITE`, `mbedtls_ssl_write` must be called again with the same arguments (same slice contents and length) until it returns `>= 0`; re-issuing with a different buffer corrupts the record stream.
- `read`: NOT cancel-safe, but it does not lose application data (a buffered transport byte is kept and consumed bytes live in the SSL context); the hazard is partial TLS input state, so it is just not side-effect-free.
- `flush` / `close`: a drop may leave a queued byte unsent or an alert partially sent.

Docs only: no behavior change, no API change, no new dependency.

## Scope note

This documents the contract; it does not try to enforce it. Making `write` cancel-detecting (e.g. an in-flight-buffer guard that rejects a resumed write with a different buffer) or genuinely cancel-safe (an internal buffer/owned-driver model) is a behavior change. A guard only detects misuse and has semantic holes (same pointer with mutated contents, same contents at a different address), and a truly cancel-safe redesign overlaps the full-duplex `split` work, so I left both as possible follow-ups rather than bundling them here. Happy to pursue either if you'd like.

## Testing

`cargo build` / `clippy` / `fmt --check` on `mbedtls-rs`, and the esp32c3 and esp32c2 example builds. The new `# Cancel safety` rustdoc intra-doc links resolve cleanly (`cargo doc` produces no new warnings from this file).

### DIFF
--- a/mbedtls-rs/src/session/asynch.rs
+++ b/mbedtls-rs/src/session/asynch.rs
@@ -120,6 +120,8 @@ where
         Ok(())
     }
 
+    // NOT cancel-safe: drives the handshake across awaits after resetting the
+    // SSL context; see `Session::connect`'s `# Cancel safety`.
     async fn connect_internal(
         &mut self,
         saved_session: Option<&SavedSession>,
@@ -142,6 +144,14 @@ where
     ///
     /// Note that calling it is not mandatory, because the TLS session is anyway
     /// negotiated during the first read or write operation, or when splitting the session.
+    ///
+    /// # Cancel safety
+    ///
+    /// NOT cancel-safe. The handshake resets the SSL context (`mbedtls_ssl_session_reset`)
+    /// before driving it across multiple `.await` points; if this future is dropped
+    /// mid-handshake, the local TLS state is left partway through a handshake the peer may
+    /// have advanced, and a retry can reset it out from under the peer.
+    // NOT cancel-safe: see `# Cancel safety`.
     pub async fn connect(&mut self) -> Result<(), SessionError> {
         self.connect_internal(None).await
     }
@@ -149,6 +159,11 @@ where
     /// Negotiate the TLS connection attempting to reuse a previously captured session.
     ///
     /// Use [`Session::save`] to get a copy of the session to use here  
+    ///
+    /// # Cancel safety
+    ///
+    /// NOT cancel-safe. Same as [`Session::connect`].
+    // NOT cancel-safe: see `# Cancel safety`.
     pub async fn connect_with_session(
         &mut self,
         saved_session: &SavedSession,
@@ -160,6 +175,13 @@ where
     ///
     /// # Returns
     /// - A tuple containing the read and write halves of the session
+    ///
+    /// # Cancel safety
+    ///
+    /// NOT cancel-safe. This negotiates the connection first (see
+    /// [`Session::connect`]); once connected the split itself has no further
+    /// `.await` points.
+    // NOT cancel-safe: see `# Cancel safety`.
     pub async fn split(
         &mut self,
     ) -> Result<
@@ -205,6 +227,15 @@ where
     ///
     /// # Returns
     /// - The number of bytes read or an error
+    ///
+    /// # Cancel safety
+    ///
+    /// NOT cancel-safe. This drives MbedTLS across multiple `.await` points. A
+    /// dropped read does not lose application data (a buffered transport byte is
+    /// kept, and bytes already consumed by MbedTLS live in the SSL context), but
+    /// it can leave partial TLS input/record state, so re-issuing is not
+    /// side-effect-free.
+    // NOT cancel-safe: see `# Cancel safety`.
     pub async fn read(&mut self, buf: &mut [u8]) -> Result<usize, SessionError> {
         self.connect().await?;
 
@@ -222,6 +253,18 @@ where
     ///
     /// # Returns:
     /// - The number of bytes written or an error
+    ///
+    /// # Cancel safety
+    ///
+    /// NOT cancel-safe. MbedTLS copies part of `data` into its record buffer, so
+    /// after an internal `WANT_WRITE` `mbedtls_ssl_write` must be called again
+    /// with the same arguments (same slice contents and length) until it returns
+    /// `>= 0`. If this future is dropped mid-write and `write` is then re-issued
+    /// with a different buffer, the peer receives the previously-buffered record
+    /// while this call reports the new data as sent, corrupting the stream. Retry
+    /// a cancelled write only with the identical buffer; do not interleave
+    /// different writes.
+    // NOT cancel-safe: see `# Cancel safety`.
     pub async fn write(&mut self, data: &[u8]) -> Result<usize, SessionError> {
         self.connect().await?;
 
@@ -238,6 +281,14 @@ where
     ///
     /// # Returns:
     /// - An error if the flush failed
+    ///
+    /// # Cancel safety
+    ///
+    /// NOT cancel-safe. A dropped flush may leave a queued transport byte unsent
+    /// or the underlying stream only partially flushed, so it is not
+    /// side-effect-free; re-flushing is generally fine if the underlying `Write`
+    /// is well-behaved.
+    // NOT cancel-safe: see `# Cancel safety`.
     pub async fn flush(&mut self) -> Result<(), SessionError> {
         self.connect().await?;
 
@@ -250,6 +301,12 @@ where
     ///
     /// # Returns:
     /// - An error if the close failed
+    ///
+    /// # Cancel safety
+    ///
+    /// NOT cancel-safe. Sends the close-notify alert and flushes; a drop may
+    /// leave the alert partially sent.
+    // NOT cancel-safe: see `# Cancel safety`.
     pub async fn close(&mut self) -> Result<(), SessionError> {
         if !self.connected {
             return Ok(());
@@ -297,6 +354,7 @@ impl<T> Read for Session<'_, T>
 where
     T: Read + Write,
 {
+    // NOT cancel-safe: forwards to `Session::read`; see its `# Cancel safety`.
     async fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
         Self::read(self, buf).await
     }
@@ -306,10 +364,12 @@ impl<T> Write for Session<'_, T>
 where
     T: Read + Write,
 {
+    // NOT cancel-safe: forwards to `Session::write`; see its `# Cancel safety`.
     async fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
         Self::write(self, buf).await
     }
 
+    // NOT cancel-safe: forwards to `Session::flush`; see its `# Cancel safety`.
     async fn flush(&mut self) -> Result<(), Self::Error> {
         Self::flush(self).await
     }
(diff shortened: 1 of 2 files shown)


ivmarkov:
> MbedTLS ssl_write must be retried with the same arguments after WANT_WRITE, so dropping a write future mid-flight and re-issuing with a different buffer makes the peer receive the previously-buffered record while the new call reports success.

I'm pretty sure this is incorrect, and `write` (and `read`) _are_ - in fact - cancel safe.
Can you push your LLM to explain in details why it believes these two are not can cancellation safe? As in a step-by-step guide through the code.

andrico21:
You're right, and thanks for pushing back. I retraced it and `read` is cancel-safe, and `write` is cancel-safe in the sense that matters: a dropped future never leaves the session in an unrecoverable state. I had the framing wrong. Here is the step-by-step.

The only thing that calls into MbedTLS is `call_mbedtls`, and its `poll_fn` always returns `Poll::Ready`. So the whole `mbedtls_ssl_write` / `mbedtls_ssl_read` runs synchronously inside a single poll and is never torn in the middle by a cancellation. That await is not a real suspension point.

The only real `.await` in each loop is `wait_writable` / `wait_readable`, which is `self.stream.write(&[byte]).await` / `self.stream.read(&mut buf).await`. If the future is dropped there, all the partial state lives in the `Session`, not in the future: the one buffered byte (`write_byte` / `read_byte`) and MbedTLS's own record state (`out_left`, `out_msg`) in the ssl context. The future only borrows `&mut self`, so dropping it strands nothing. `write_byte.take()` also runs only after the await resolves, with no cancellation point in between, so a drop leaves the byte in place and the next call resumes from exactly where it left off.

The one caveat I'd still document is not a cancel-safety problem per se: after an internal `WANT_WRITE`, MbedTLS requires the retry to use the same bytes. `ssl_write_real` flushes the already-buffered record and ignores `buf` when `out_left != 0`. So if someone cancels a pending write and then issues a different buffer, MbedTLS finishes sending the old record but reports the new length. That is the standard "retry the same bytes" `Write` contract, not session corruption.

I'll push a commit that re-scopes the docs accordingly: `read` and `write` documented as cancel-safe, with a short note on `write` that an interrupted write must be resumed with the same data. Does that match how you intend the contract, or would you rather I just drop the cancel-safety docs entirely?

ivmarkov:
> The one caveat I'd still document is not a cancel-safety problem per se: after an internal WANT_WRITE, MbedTLS requires the retry to use the same bytes. ssl_write_real flushes the already-buffered record and ignores buf when out_left != 0. So if someone cancels a pending write and then issues a different buffer, MbedTLS finishes sending the old record but reports the new length. That is the standard "retry the same bytes" Write contract, not session corruption.

I do not understand this, can you elaborate?

Typical `Write::write` contract is as follows:
1. Cancellation-safe
2. If cancelled then you have a guarantee that the passed data was NOT send. **Not even a single byte**
3. If NOT cancelled, then exactly `usize` number bytes of the passed data was sent, where this `len` is returned by the `write` method. Therefore, on the next `Write::write` call, you have to anyway **not** pass the first `usize` bytes from the `data: &[u8]`. Now, whether - on the second call - you pass precisely the remaining `&data[len..]` or not is up to you - the `write` interface does not care actually. Just as it does not care in case you are not careful and pass again the complete `data: &[u8]` without accounting for the bytes already sent

So what am I missing?
My point is, if we have a gap and we do not faithfully implement ^^^, this need to be fixed.

On the other hand, if you are willing to document details of the "typical Write contract" on the mbedtls-rs methods implementing that contract, I don't see the point? This should be documented on the `Write` trait level in `embedded_io(_async)`?

ivmarkov:
> So if someone cancels a pending write and then issues a different buffer, MbedTLS finishes sending the old record but reports the new length

This is where I'm struggling.

andrico21:
You are right, my "standard retry-same-bytes contract" line was a hand-wave that misses a real bug.

Within one `write(data)` future the loop always re-calls `mbedtls_ssl_write` with the same `data`, so MbedTLS's "same parameters after WANT_WRITE" holds. The gap is across calls. If `write(A)` is dropped at `wait_writable` (the only real `.await`), MbedTLS keeps A's pending record (`out_left != 0`) and we keep the buffered byte in `write_byte`, both in the `Session`. A later `write(B)` calls `mbedtls_ssl_write(B)`; MbedTLS sees `out_left != 0`, flushes A's record, ignores `B`, returns `B.len()`. So `write(B)` returns `Ok(n)` while A's bytes went on the wire and none of B did. That breaks your point 3, and it is a code bug, not a docs matter.

Point 2 is the weaker one: cancel-safety is encouraged not required, and a cancelled write is genuinely not side-effect-free (part of A may already be on the wire), so that part is just a "not cancel-safe on writes" note.

So #152 shouldn't land as docs-only. Either convert it to a bugfix (the accounting issue, plus the cancel note) or drop it for a separate bugfix PR, your call. The fix property is "never hand a new buffer to `mbedtls_ssl_write` while old output is pending": after a cancelled in-flight write, drain the old record first without counting it against the new buffer, or fail the next write. Transparently resuming the old write and returning its real count is not possible without owning the buffer across calls, which the `&[u8]` API doesn't. I'll write up the minimal fix once you pick a direction.

andrico21:
Here's the fix I'd propose for the accounting bug.

Scope: async only. The blocking `write` loops `WANT_WRITE` synchronously with no future to drop, so it can't hit this.

Plan:
- Add a small write-side "drain pending" guard on the async `Session`, set right before the `wait_writable().await` and cleared on every outcome of that await. If it survives into the next `write()`, the previous write was cancelled mid-record (MbedTLS still has `out_left != 0`).
- On the next `write(data)`, if the guard is set, first finish the old record with `mbedtls_ssl_flush_output` (not a zero-length `mbedtls_ssl_write`, since the public write runs renegotiation/handshake checks before `ssl_write_real`), then start a fresh record for `data` and return only `data`'s count.

That restores point 3 (a write reports only its own bytes). It does not make writes side-effect-free on cancel (a cancelled write may still put its record on the wire), so I'd document the write side as not cancel-safe, and leave `read` as is since it never misreports.

No public API change, no change to blocking. Does this approach look right to you, and do you want it as a new commit on this PR or a separate bugfix PR?

ivmarkov:
I was looking at this myself, and I agree `write` is **NOT** cancel safe. (`read` seemingly is).

So yes, the minimum would be to (a) document this and (b) do some sort of "drain pending" guard as suggested above.

If we want to recover the full cancellation-safety of `write` though, this can only be done with an extra buffer. Would be also interesting to see such a wrapper - `BufferedWriteSession<S: Session>`, which wraps a `Session` and a `BufWriter` from `embedded_io_async` (I think there was one).

The `BufferedWriteSession` would expose the same API as `Session`, with the one big difference that `write` would be cancellation-safe.

And then the user can pick and choose if they want true cancellation-safety for `write` (= pay extra price for a buffer), or not.

ivmarkov:
Related note: I think ideally we should call `mbedtls_ssl_flush_output` from the `flush` method as well, and we don't do that?

andrico21:
Sounds good, I'll go with (a) document + (b) the drain-pending guard.

On `flush`: you're right, and we don't. Both the async and blocking `flush` only flush the underlying stream (the async one also pushes the single byte staged by `wait_writable`), neither calls `mbedtls_ssl_flush_output`, so a partial record left in MbedTLS (`out_left != 0`) wouldn't be pushed by `flush`. It's latent today because `write` normally runs `ssl_write` to completion so `out_left` is 0 by the time you flush, but semantically `flush` should guarantee MbedTLS's pending output reaches the transport. I'll add a `mbedtls_ssl_flush_output` call to both `flush` impls. It's the same primitive the drain-guard uses, so it fits naturally in this PR.

On `BufferedWriteSession<S>`: nice idea. `embedded_io_async` does have `BufWriter`, so a wrapper that owns a buffer and re-feeds the same bytes after a cancel would give true write cancellation-safety, opt-in for the extra RAM. I'd rather land the accounting fix + the `flush` fix here first and do `BufferedWriteSession` as a follow-up, if that works for you.

So this PR becomes: drain-pending guard so a cancelled write never misreports, `flush` actually flushes MbedTLS output, and the cancel-safety docs (write not side-effect-free, read fine). Want it on this PR, or split the `flush` fix out?

ivmarkov:
> So this PR becomes: drain-pending guard so a cancelled write never misreports, `flush` actually flushes MbedTLS output, and the cancel-safety docs (write not side-effect-free, read fine). Want it on this PR, or split the `flush` fix out?

This PR - thanks!

andrico21:
Small correction to my last comment: `mbedtls_ssl_flush_output` turns out to be internal (not in the public headers or our bindings), so I can't call it directly. The public way to flush a pending record is to re-enter `mbedtls_ssl_write`, so the fix uses a guarded zero-length `mbedtls_ssl_write` that only runs when a record is actually pending (an unguarded zero-length write would emit an empty TLS record, per the `ssl.h` note).

Also, this is async-only: the blocking BIO never returns `WANT_WRITE` (a full or failed transport surfaces as `CONN_EOF`), so blocking `write`/`flush` drain synchronously within the call and don't need the guard. So the change lands only in the async `write`/`flush`/`close`.

ivmarkov:
> Small correction to my last comment: `mbedtls_ssl_flush_output` turns out to be internal (not in the public headers or our bindings), so I can't call it directly. The public way to flush a pending record is to re-enter `mbedtls_ssl_write`, so the fix uses a guarded zero-length `mbedtls_ssl_write` that only runs when a record is actually pending (an unguarded zero-length write would emit an empty TLS record, per the `ssl.h` note).
> 
> Also, this is async-only: the blocking BIO never returns `WANT_WRITE` (a full or failed transport surfaces as `CONN_EOF`), so blocking `write`/`flush` drain synchronously within the call and don't need the guard. So the change lands only in the async `write`/`flush`/`close`.

That's OK.

mbedtls-rs/src/session/asynch.rs
```diff
@@ -370,12 +439,20 @@ where
...
 
 impl<T> SessionRead<'_, T>
 where
     T: Read,
 {
+    /// Read unencrypted data from the read half of the TLS connection.
+    ///
+    /// # Cancel safety
+    ///
+    /// NOT cancel-safe. Same as [`Session::read`].
```
ivmarkov: I think read **IS** cancel-safe. The _other_ read isn't because it calls `connect`. But this one should be cancel-safe.

ivmarkov (commented):
LGTM except one comment.

andrico21:
You're right, fixed.

`SessionRead::read` now documents `# Cancel safety` as cancel-safe, with the rationale (handshake has already completed by the time `Session::split` could produce this half, only drives MbedTLS's data-read loop, partial state lives on the half itself not in the future, contrast with `Session::read` which inherits the connect-handshake hazard). The `impl Read` forwarder gets the same flip.

I also tightened the `MBio::read` primitive's comment in passing, since it was claiming "NOT cancel-safe" by reference to `Session::read`'s docs — but `MBio::read` itself is cancel-safe (its only `.await` is `wait_readable`, whose partial state lives on the read half, not in the future); `Session::read`'s unsafety comes from the `connect()` call wrapping it, not from this primitive. Easy to revert that one in particular if you'd rather keep the doc scope minimal.

PR #155, opened 2026-06-16
Bind a saved session to its server name and reject cross-host resume

andrico21:
Hi Ivan, another finding from the same audit, independent of the others.

## What

`SavedSession` carries no peer binding: `connect_with_session` feeds it to any new session via `mbedtls_ssl_set_session` regardless of server name. On **TLS 1.2** a resumed handshake jumps straight to the change-cipher-spec state and **skips the certificate parse/verify states**, so a session saved for host A presented to a server B that accepts it (shared resumption secret / same infra) resumes and **skips validating B's certificate**. TLS 1.3 already rejects a ticket-hostname mismatch in C; TLS 1.2 does not.

## Fix

Track the configured server name in the session and store it in `SavedSession`, then reject a mismatch with `MBEDTLS_ERR_SSL_BAD_INPUT_DATA` **before** installing the saved session, in both the async and blocking `connect_internal`. The check covers both protocol versions (defence in depth; 1.3 already enforces in C). `set_server_name` is also rejected after the handshake, so a saved session can't be bound to a name that wasn't used to negotiate it.

Reuses the existing error path rather than adding a `SessionError` variant. `SavedSession` gains a private field, so it's not a Rust source break.

## Behaviour change

- A cross-host `connect_with_session` that previously resumed now returns an error (the point of the fix).
- `set_server_name` after the handshake now errors.
- A session saved without a server name carries no peer binding and may only be resumed into an equally nameless session.

Adds `saved_session_binding.rs` locking the comparison contract.

### DIFF
--- a/mbedtls-rs/src/session.rs
+++ b/mbedtls-rs/src/session.rs
@@ -15,9 +15,96 @@ pub use asynch::*;
 mod asynch;
 pub mod blocking;
 
+/// An owned, nul-terminated server name allocated through MbedTLS's allocator
+/// (`mbedtls_calloc`/`mbedtls_free`), so it honours a user-overridden MbedTLS
+/// heap instead of the Rust global allocator. Used to bind a saved session to a
+/// peer identity (see [`SavedSession`]).
+pub(crate) struct ServerName {
+    ptr: NonNull<u8>,
+}
+
+impl ServerName {
+    /// Copy a `&CStr`'s bytes (including the nul) into a fresh MbedTLS-allocated
+    /// buffer. Returns `None` if the allocation fails.
+    fn from_cstr(name: &CStr) -> Option<Self> {
+        Self::from_bytes_with_nul(name.to_bytes_with_nul())
+    }
+
+    fn from_bytes_with_nul(bytes: &[u8]) -> Option<Self> {
+        // SAFETY: `mbedtls_calloc` is the MbedTLS allocator entry point; requesting
+        // `bytes.len()` elements of one byte each yields a `bytes.len()`-byte (or
+        // null) allocation, which `NonNull::new` checks before we treat it as owned.
+        let ptr =
+            NonNull::new(unsafe { mbedtls_calloc(bytes.len(), size_of::<u8>()) }.cast::<u8>())?;
+
+        // SAFETY: `ptr` points to a fresh `bytes.len()`-byte allocation, and the
+        // source slice is independent of it.
+        unsafe {
+            core::ptr::copy_nonoverlapping(bytes.as_ptr(), ptr.as_ptr(), bytes.len());
+        }
+
+        Some(Self { ptr })
+    }
+
+    fn as_bytes(&self) -> &[u8] {
+        // SAFETY: `ptr` was constructed from a nul-terminated source slice in
+        // `from_bytes_with_nul` and is never mutated after construction.
+        unsafe { CStr::from_ptr(self.ptr.as_ptr().cast::<c_char>()) }.to_bytes_with_nul()
+    }
+}
+
+impl Drop for ServerName {
+    fn drop(&mut self) {
+        // SAFETY: `ptr` was allocated by `mbedtls_calloc` in `from_bytes_with_nul`,
+        // is owned solely by this `ServerName`, and is freed exactly once here.
+        unsafe {
+            mbedtls_free(self.ptr.as_ptr() as *mut c_void);
+        }
+    }
+}
+
 /// A reusable TLS session state captured from a connected session.
+///
+/// A saved session is bound to the server name that was configured when it was
+/// captured, and [`Session::connect_with_session`] refuses to resume it against
+/// a different server name (see that method). A session captured without a
+/// server name carries no peer binding and may only be resumed into an equally
+/// nameless session.
 pub struct SavedSession {
     pub(crate) mbedtls_session: MBox<mbedtls_ssl_session>,
+    /// The server name the originating session was configured with, used to
+    /// reject cross-host resume. `None` if the session had no server name.
+    pub(crate) server_name: Option<ServerName>,
+}
+
+/// Reject reusing a saved session against a different peer identity.
+///
+/// TLS 1.2 resume acceptance does not re-validate the server certificate, so a
+/// session saved for host A presented to a server B that accepts it would skip
+/// verifying B's certificate. We enforce the binding in the wrapper for both TLS
+/// 1.2 and 1.3 (1.3 also checks in C). `None`/`Some` are treated as distinct, so
+/// a nameless saved session only resumes into a nameless session.
+pub(crate) fn check_saved_session_server_name(
+    saved: &Option<ServerName>,
+    current: *const c_char,
+) -> Result<(), SessionError> {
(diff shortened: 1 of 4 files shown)


mbedtls-rs/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 #![no_std]
 #![allow(clippy::uninlined_format_args)]
 
+extern crate alloc;
```
ivmarkov: This is NOK.

While MbedTLS does need a C allocator (`calloc` + `free`), and one of the ways to provide those is via the Rust heap (see `tinyrlibc` `calloc/free`), this is just _one_ of the ways to do this.

User can override the `calloc` & `free` functions used by MbedTLS, and OpenThread - for one - typically does that so as to redirect MbedTLS allocations to a small buffer-based heap.

Therefore, if you absolutely need to use "Heap" memory, Do it with an `MBox`.

ivmarkov (changes requested):
Direct usage of the Rust heap in `MbedTLS` is not ideal.

andrico21:
Good catch, and agreed. Routing this through the Rust global heap was wrong precisely because it bypasses the user-overridable MbedTLS allocator (the OpenThread small-buffer-heap case you describe is exactly the scenario it would break).

I've reworked it so the only server-name allocation now goes through `mbedtls_calloc`/`mbedtls_free`, and dropped `extern crate alloc` entirely.

One note on `MBox` specifically: it wraps a single `T: MInit` value, so it can't own a variable-length `[u8]` name buffer. Instead I added a small RAII `ServerName` type that mirrors the existing `ALPNArray` in `session.rs` (same `mbedtls_calloc` + `Drop`-time `mbedtls_free` pattern already used in this crate for the ALPN list). It stores the nul-terminated name, and the saved-session deep-copy is fallible: if the MbedTLS allocator is out of memory it returns `MBEDTLS_ERR_SSL_ALLOC_FAILED` rather than silently dropping the name, so a named session can never degrade to a nameless one (which would weaken the binding check).

`extern crate alloc` is gone and there's no remaining `Vec`/Rust-heap use in these files. Pushed onto the same commit; the esp32c2/esp32c3 no-std builds and the rest of the CI matrix are green.

mbedtls-rs/src/session.rs
```diff
@@ -15,9 +15,119 @@ pub use asynch::*;
...
 pub mod blocking;
 
+/// An owned, nul-terminated server name allocated through MbedTLS's allocator
+/// (`mbedtls_calloc`/`mbedtls_free`), so it honours a user-overridden MbedTLS
+/// heap instead of the Rust global allocator. Used to bind a saved session to a
+/// peer identity (see [`SavedSession`]).
+pub(crate) struct ServerName {
+    ptr: NonNull<u8>,
+    /// Length in bytes, including the trailing nul.
+    len: usize,
```
ivmarkov: Given that `ptr` is always a null-terminated string, this field is unnecessary.

mbedtls-rs/src/session.rs
```diff
@@ -251,6 +361,9 @@ struct SessionState<'a> {
     /// While not explicitly used, we need to keep a reference to it as it is used
     /// by the SSL context via a raw pointer
     _alpn_ptrs: Option<ALPNArray<'a>>,
+    /// The configured server name, tracked so a saved session can be bound to it
+    /// and cross-host resume rejected. See [`check_saved_session_server_name`].
+    server_name: Option<ServerName>,
```
ivmarkov: I do not understand why we need this state. 
Isn't it possible to use `mbedtls_ssl_get_hostname` instead? The thing is, the SSL context is _anyway_ keeping a copy of the same state (the server name) so why duplicate it here?

ivmarkov (commented):
Sorry just spotted a couple of extra things.

andrico21:
Both fair. Going through them:

(A) The `len` field is genuinely redundant. Will drop it and derive the length lazily via `CStr::from_ptr(self.ptr.as_ptr().cast()).to_bytes_with_nul()` at the only two call sites (the byte-compare and the saved-session deep copy). One strlen per call, both amortized over a TLS handshake.

(B) Agree the live-session duplication is unjustified. The only obstacle is that `mbedtls_ssl_get_hostname` is `static inline` in `ssl.h`, so bindgen generates no FFI symbol; calling it from Rust needs a port. Concretely I'd like to add a small `mbedtls-rs-sys/src/ssl_inline.rs` module containing just a Rust port of the inline (one field read of `private_hostname`, gated on `feature = "x509-parse"` to mirror `MBEDTLS_X509_CRT_PARSE_C`), re-exported from the crate root so it's callable as `mbedtls_rs_sys::mbedtls_ssl_get_hostname`. session.rs's existing pattern (`private_min_tls_version`, `private_key.0` on lines 387/400) already reaches into the same private namespace, so this is consistent.

That lets me drop `SessionState::server_name` entirely. `SavedSession` still owns its own copy (the saved session outlives the SSL context that produced it), but the only allocation in the redesign is the saved-session copy, made via `mbedtls_calloc`/`mbedtls_free` as before, with allocation failure surfacing as `MBEDTLS_ERR_SSL_ALLOC_FAILED` so a named session never silently degrades to nameless.

OK to proceed with this approach, or would you prefer something else (e.g. a non-inline C shim instead of the Rust port, or just reading `(*ssl).private_hostname` directly in session.rs without touching mbedtls-rs-sys)?

ivmarkov:
> (B) Agree the live-session duplication is unjustified. The only obstacle is that `mbedtls_ssl_get_hostname` is `static inline` in `ssl.h`, so bindgen generates no FFI symbol; calling it from Rust needs a port. Concretely I'd like to add a small `mbedtls-rs-sys/src/ssl_inline.rs` module containing just a Rust port of the inline (one field read of `private_hostname`, gated on `feature = "x509-parse"` to mirror `MBEDTLS_X509_CRT_PARSE_C`), re-exported from the crate root so it's callable as `mbedtls_rs_sys::mbedtls_ssl_get_hostname`. session.rs's existing pattern (`private_min_tls_version`, `private_key.0` on lines 387/400) already reaches into the same private namespace, so this is consistent.

That's a bit too much for getting access to `mbedtls_ssl_get_hostname`. 
I would instead suggest accessing the  "private" field in the SSL struct**directly** - `private_hostname`?

andrico21:
Done, pushed.

Single-crate v2: dropped `ServerName::len` and `SessionState::server_name`, removed `clone_server_name`, now reading `(*ssl).private_hostname` directly in `connect_internal` and `save()` for both the async and blocking paths (same pattern session.rs already uses for `private_min_tls_version` and `private_key.0` on lines 387/400). `SavedSession` still owns its own MbedTLS-allocator-backed copy because it outlives the SSL context; allocation failure on that copy still surfaces as `MBEDTLS_ERR_SSL_ALLOC_FAILED` so a named session never silently degrades to nameless.

Went with your suggestion since the diff is smaller and the pattern is already in place. The alternative I'd been planning was a tiny `mbedtls-rs-sys::ssl_inline` module wrapping the static-inline as a documented `mbedtls_ssl_get_hostname` accessor; happy to add that as a follow-up if you'd ever prefer the named helper, but no pressure either way.

PR #158, opened 2026-06-22
Validate xtask target triple before deleting libs/<target> (F-49)

andrico21:
`xtask gen` takes the Rust target triple as a free-form CLI string and
joins it onto `mbedtls-rs-sys/libs/<target>` and
`mbedtls-rs-sys/src/include/<target>.rs`. The first is later passed to
`fs::remove_dir_all` in `harvest` to clear stale artifacts. With no
validation, `xtask gen ../../etc` resolved `libs/../../etc` to `etc/`
relative to CWD and called `remove_dir_all` on it, and `xtask gen
/etc/passwd` made `Path::join` discard the prefix (absolute path
replaces) and called `remove_dir_all("/etc/passwd")`. A dev tool, so
low blast radius, but trivial to fix.

Add `validate_target_triple` and call it on the parsed `target` before
any path is built from it. Each `-`-separated segment must match
`[a-z0-9_.]+` and start with `[a-z0-9_]` (no leading dot); the
substring `..` is rejected anywhere; the whole string must have at
least one `-` and be at most 64 chars. The `.` is permitted in any
segment so real Armv8-M targets (`thumbv8m.main-none-eabi`,
`thumbv8m.main-none-eabihf`, `thumbv8m.base-none-eabi`) continue to
work; `..` is still rejected so traversal cannot smuggle through a
dotted segment.

As belt-and-braces, the libs root is canonicalized before the later
`remove_dir_all` and `ensure!` asserts the joined target still lives
under it. `ensure!` (not `debug_assert!`) so the check survives
release builds, since `xtask` is typically run via `cargo run
--release`.

Inline tests cover every real target the upstream CI matrix builds
(`x86_64-unknown-linux-gnu`, `aarch64-apple-darwin`, all the riscv32
/ xtensa / thumbv7 / thumbv8m / wasm32-wasip1 leaves) plus 18
rejection cases (`..` traversal, absolute paths, `/` and `\`
separators, uppercase, spaces, control chars, leading-dash,
trailing-dash, single-segment, empty, and over-length). Manual smoke
verified: `xtask gen ../../etc` and `xtask gen /etc/passwd` both
exit non-zero with a clear `invalid target triple` message before
any filesystem mutation; `xtask gen thumbv8m.main-none-eabi`
proceeds into the existing build path unchanged.

Refs: CODE_REVIEW.md sections 249 and 415.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -4,7 +4,7 @@ use std::io::{BufRead, BufReader};
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
-use anyhow::{anyhow, bail, Context, Result};
+use anyhow::{anyhow, bail, ensure, Context, Result};
 
 use clap::{Parser, Subcommand};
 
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -347,3 +373,110 @@ fn harvest(out_dir: &Path, libs_dst: &Path, bindings_dst: &Path) -> Result<()> {
 
     Ok(())
 }
+
+/// Validates that `target` is a well-formed Rust target triple. Each `-`
```
ivmarkov, line 377: Same comments here: F-49 etc.

ivmarkov (commented):
Well, this PR will not harm  (I guess) but I wonder why we need to go at these great lengths?

Sure, an invalid target triple might delete the wrong stuff, but in the end - this is all human-controlled and the target tripple does not come via a REST request or anything like that?

Regardless - if you fix the comments I'll merge.
Also good to replicate this for `openthread`.

andrico21:
Pushed comment cleanup at `baf29c9`, ready for re-review.

On openthread: looked at `esp-rs/openthread/xtask/src/main.rs` - same pattern (line 58 `target: String`, line 95 `.join("libs").join(&target)`, line 238 `fs::remove_dir_all`). I'll port the same validator once this one lands.

PR #159, opened 2026-06-22
Validate codegen inputs in the user-config builder (F-29)

andrico21:
`mbedtls-rs-sys/gen/config.rs` exposes `MbedtlsUserConfig::set` and
`::add_include`, which render their arguments verbatim into C preprocessor
directives via the `Display` impl (`#undef MBEDTLS_{ident}`, `#define
MBEDTLS_{ident} {literal}`, `#include "{header}"`). Every caller today is in
`gen/features.rs` and `gen/builder.rs` and passes a compile-time string
constant, so the templates are safe in practice. The API is `pub`, though,
so a future caller fed config or env input could inject arbitrary
`#define` / `#include` / comment-token directives or escape the quoted
include.

Tighten the surface and validate every byte at insertion time. `pub`
becomes `pub(crate)` on `MbedtlsUserConfig`, `Value`, and every inherent
method. The whole `gen/` tree is reachable only via the `#[path]` import
in `build.rs`, so the entire builder type was already unreachable from
`lib.rs`; tightening is non-breaking. `set` runs `validate_macro_ident`
(non-empty, `^[A-Z0-9_]+$`) on the identifier and, when the value is a
literal, `validate_macro_literal` (no NUL / `\n` / `\r` / `"` / `\`, no
`/*` / `*/` / `//`, no `??` C trigraph sequences, covering line-escape,
comment-token smuggling, and trigraph-backed line continuation under any
toolchain mode that honours trigraphs). `add_include` runs
`validate_header_path`, a strict `matches!` allowlist of known-good
MbedTLS user headers; adding a new hook header now requires an explicit
edit to the allowlist, making every new include reviewer-visible. The
allowlist doc cross-references why `sha{1,256,512}_alt.h` are absent
(they reach MbedTLS via the `-I gen/hook/` flag automatically; only
`time_alt.h` needs an explicit user-config `#include`). The validators
`assert!` because every existing caller passes a compile-time string
constant; a violation indicates a codegen bug, not a runtime user-input
failure.

Tests live in a `#[cfg(test)] mod tests` block in `gen/config.rs` and
are exercised through `mbedtls-rs-sys/tests/config_validators.rs`, which
`#[path]`-loads the module so `pub(crate)` becomes test-crate-local. 28
cases cover positive accepts (uppercase ident, digits, underscores,
literal values, allowlisted header, render round-trip) and rejections
across identifier shape, literal injection (newline, quote, backslash,
NUL, comment tokens, carriage return, trigraph backslash / hash / pipe),
and header allowlist escape attempts.

Refs: CODE_REVIEW.md sections 220 and 395.

### DIFF
--- a/mbedtls-rs-sys/gen/config.rs
+++ b/mbedtls-rs-sys/gen/config.rs
@@ -7,30 +7,43 @@ use std::{fmt, fs, io};
 ///
 /// The user config is not the full MbedTLS config, it's appended to the
 /// default config.
+///
+/// This type and its inherent methods are intentionally `pub(crate)`: the
+/// `gen/` tree is only reachable via the `#[path]` import in
+/// `mbedtls-rs-sys/build.rs`, so there is no downstream API surface to
+/// preserve and tightening visibility eliminates any future call site that
+/// could feed unvalidated strings into the C preprocessor template rendered
+/// by `Display::fmt`.
 #[derive(Debug)]
-pub struct MbedtlsUserConfig {
+pub(crate) struct MbedtlsUserConfig {
     options: BTreeMap<Box<str>, Value>,
     includes: Vec<Box<str>>,
 }
 
 impl MbedtlsUserConfig {
     /// Creates a new, empty config.
-    pub fn new() -> Self {
+    pub(crate) fn new() -> Self {
         Self {
             options: BTreeMap::new(),
             includes: Vec::new(),
         }
     }
 
-    /// Appends a `#include` directive
+    /// Appends a `#include` directive.
     ///
-    pub fn add_include(&mut self, header: &str) -> &mut Self {
+    /// `header` is rendered verbatim into `#include "{header}"`. To keep
+    /// that template injection-safe, the header name must appear in a
+    /// strict allowlist; adding a new MbedTLS user header requires an
+    /// explicit edit to [`validate_header_path`]. Asserts (panics on bug)
+    /// because all callers in `gen/` pass compile-time string constants.
+    pub(crate) fn add_include(&mut self, header: &str) -> &mut Self {
+        validate_header_path(header);
         self.includes.push(header.into());
         self
     }
 
     /// Writes the MbedTLS user config to a header file.
-    pub fn write_to_path(&self, path: &Path) -> io::Result<()> {
+    pub(crate) fn write_to_path(&self, path: &Path) -> io::Result<()> {
         let contents = self.to_string();
         fs::write(path, contents)
     }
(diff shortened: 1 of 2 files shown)


mbedtls-rs-sys/gen/config.rs
```diff
@@ -161,3 +187,280 @@ impl From<String> for Value {
...
+    assert!(
+        !lit.contains("??"),
+        "BUG: invalid MbedTLS config literal {lit:?}: must not contain `??` (trigraph prefix is forbidden)",
+    );
+}
+
+/// Header path validator. Strict allowlist of known-good MbedTLS user headers
+/// per CODE_REVIEW.md section 10 F-29 ("allowlist includes"). Add new entries
+/// here when (and only when) a new hook requires them; the explicit edit is
+/// the point: every new include is reviewer-visible.
```
ivmarkov: Can you remove from the comments all references to artefacts which are not part of this repo? Like CODE_REVIEW.md. "F-29". and so on.

Also in general we don't need - in the comments - the "history" as to how this code came to be.

mbedtls-rs-sys/gen/config.rs
```diff
@@ -7,30 +7,44 @@ use std::{fmt, fs, io};
...
             options: BTreeMap::new(),
             includes: Vec::new(),
         }
     }
 
-    /// Appends a `#include` directive
+    /// Appends a `#include` directive.
     ///
-    pub fn add_include(&mut self, header: &str) -> &mut Self {
+    /// `header` is rendered verbatim into `#include "{header}"`. To keep
```
ivmarkov, line 34: Ditto

mbedtls-rs-sys/tests/config_validators.rs
```diff
@@ -0,0 +1,21 @@
+//! Integration test entry point for the codegen input validators in
```
ivmarkov, line 1: Ditto

ivmarkov (commented):
Feedback regarding the comments only. I've noticed that Opus tends to over-comment, and often puts unnecessary stuff in the comments (these are abused as its own memory or "line of thought").

Also - would be good to have a similar PR for `openthread`.

andrico21:
Pushed comment cleanup at `f74d036`, ready for re-review.

On openthread: looked at `openthread-sys/gen/builder.rs` - it uses `cmake::Config::define(key, val)` directly rather than rendering its own `#define MBEDTLS_{ident}` template, so this PR's validation mechanism doesn't have a 1:1 analog there. The dynamic `define()` callsite at line 318 (`config.define(setting.knob, ...)`) is potentially a similar surface but the validation shape would be different. Can revisit if you see a specific path that wants tightening - say which.

PR #164, opened 2026-07-05
AES and ECC accel for esp32XX

ivmarkov:
This PR adds the missing AES and Elliptic Curves hw accel for the ESP MCUs.

Additionally - and because the used accelerators from `esp-hal` became quite a few (sha1, sha256, sha512, exp-mod, ecc, aes), The `EspAccel` type now has a builder-style API that allows the user to only instantiate _some_ of the hw accel backends, if so desired.

Even more importantly, `mbedtls-rs` is no longer opinionated as to who has to instantiate the `esp-hal` backends and their queues. While the user might use the `EspAccel` type for this, the user could also do this manually, using direct calls into `esp-hal`.

In other words, hooking an algorithm into the ESP HW accel framework is now decoupled (the `EspHooks` type) from instantiating the HW accel backends and their queues (`EspAccel`).

Last but not least, the alignment of all work areas now uses `addr_of_mut!` instead of `&mut` access. 
Reason: even though we request alignment=16, heap allocators like the OpenThread internal allocator, only aligns at the 8-byte boundary. Note that:
- `alignment=16` is necessary when the context struct is allocated on-stack, due to Rust move semantics
- When the context struct is allocated on-C-heap by an allocator with smaller alignment support, we are still ok, as we "re-align" inside the work area the actual Rust payload. It is just that this re-alignment cannot be done when Rust `memmove`s stuff around with on-stack unpinned types.


ivmarkov:
**UPDATE:** We now dedicated **P256-Sign** and **P256-Verify** tests that try to avoid the "software" bias of the ECP test and indeed show much bigger accel with the ESP ECC hardware.

============

#### Why does the ESP ECC hardware acceleration show only ~1.85× over the software fallback?

(Answered by Fable.)

> **Q:** The ECP self-test shows a very modest 1.85× speedup (less than even 2×) compared to the software fallback. Why could that be?

A 1.85× is actually within the realistic envelope for this particular benchmark on this particular peripheral — the number is the product of four compounding factors, ordered by weight:

**1. The ESP ECC engine is intrinsically modest.** It's a compact sequential modular-arithmetic block (one multiplier working through the double-and-add ladder at peripheral clock), designed primarily for *offload and constant-time hardening*, not raw throughput. A point multiplication still takes on the order of a millisecond in hardware, while mbedtls's software path on a 160 MHz RV32IMAC with an optimized comb method is only a few milliseconds for P-192. Espressif's own published mbedtls benchmarks for these chips show similarly small single-digit ratios for ECC — this is not an AES-style 10-50× peripheral.

**2. The self test is tilted *toward* software — comb-table caching.** Test #1 does repeated **base-point** multiplications with the *same reused group object*, and mbedtls caches the comb precomputation table in `grp->T` after the first mul — so software pays the precompute once and then runs cheap cached muls for the rest of the loop. The hardware path pays its full fixed cost on every single call. Since the reported figure averages over the whole test, the software side gets a structural discount that no production workload (fresh handshakes) would give it.

**3. Fixed per-call plumbing on the hook path.** Each hooked mul does MPI→byte-array conversion, a work-queue post, an interrupt-completion round trip through the executor, and bytes→MPI back. That's a roughly constant tax per operation — negligible against a 100 ms operation, material against a ~1-2 ms one.

**4. P-192 minimizes the ratio.** The self test prefers secp192r1; the smaller field makes *software* disproportionately faster (fewer, smaller limb operations), while the hardware cost and the plumbing tax barely shrink. On P-256 the same comparison would look better.

Note what the number does confirm: it's meaningfully above 1.0×, so the hardware path *is* engaged (a silent fallback would benchmark at ~1×).

If you want a figure that reflects what the accel is actually worth in production, the self test is the wrong instrument — the honest benchmark is **ECDSA P-256 sign/verify** (fresh group per operation, no comb cache, the field size TLS/Matter actually uses, and verify does the double-mul where hardware helps most). That's a small addition to the `crypto_self_test` example — a timed loop over `mbedtls_ecdsa_sign`/`verify` with a generated key — rather than anything in the sys crate.

ivmarkov:
@bjoernQ This is ready for review.

The [OpenThread Spinel (Remote radio) support](https://github.com/esp-rs/openthread/pull/98) depends on the `addr_of_mut!` alignment hardening done in this PR.

bjoernQ (approved):

PR #166, opened 2026-07-07
Fix the crate build when defmt is enabled

ivmarkov:
Subject says it all. We should use `core::assert` instead of `defmt::assert` as these asserts are actually compile-time errors.


ivmarkov:
@bjoernQ Merging without approval because the fix is really trivial.

PR #168, opened 2026-07-08
Enforce the short-enums policy across all of clang, GCC and bindgen

ivmarkov:
ARM Cortex M issue but the fix is generic:

When we migrated from compiling with GCC by default to compiling with Clang by default, we used the default short-enums policy of Clang, which is `short-enums = false`, but we DID kept passing unconditionally `short-enums = true` **to bindgen** (because GCC is `short-enums = true` by default, and we used to build with GCC).

The fact that GCC and Clang can't agree on what should be the default for short-enums is not nice, as it hinders interop between libraries compiled with different compilers.

Seemingly, the consensus is that regardless of compiler defaults, libs should be compiled with `short-enums = true` for Cortex M, and this is what this PR does by enforcing the `short-enums() fn call` outcome everywhere regardless of compiler defaults (the `short_enums()` fn simply returns `true` for Cortex, and `false` for everything else) .


ivmarkov:
@bjoernQ - Ditto here - merging this one-liner.

PR #170, opened 2026-08-20
Add cooperative restartable-ECC support to TLS sessions

scottmcnab:
## Summary

This adds opt-in support for Mbed TLS restartable elliptic-curve operations to the TLS session wrappers.

When `ecp-restartable` is enabled and Mbed TLS returns `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS`, the common async invocation helper now self-wakes, returns `Poll::Pending`, and resumes the same Mbed TLS operation on the next poll. This lets a bare-metal async executor run other work between bounded ECC steps. The blocking session's common invocation helper re-enters an in-progress operation until it completes, invoking a platform yield hook (`Session::new_with_yield`, e.g. `std::thread::yield_now`) between the bounded slices — both session types complete restartable operations transparently.

The feature is opt-in at both levels:

- Cargo enables `MBEDTLS_ECP_RESTARTABLE` together with the ECP core it requires. Mbed TLS's configuration checks require at least one elliptic curve whenever ECP is compiled, so the consumer enables its own `curve-*` selection alongside the feature.
- The application chooses the process-wide operation budget with `ecp::set_restartable_max_ops`.

Existing configurations and session behavior remain unchanged when the feature is disabled.

## Changes

- Add the `ecp-restartable` feature to `mbedtls-rs-sys` and `mbedtls-rs`.
- Ensure the changed C context layout invalidates incompatible prebuilt libraries and bindings, causing an on-the-fly source build.
- Handle `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS` in each session type's common Mbed TLS invocation helper, so every current and future wrapper operation (connect, read, write, close) gets the same semantics: the async helper self-wakes and yields with `Poll::Pending`, and the blocking helper re-enters the operation until it completes. The handling is unconditional (not gated on the Cargo feature): the error code is always defined and never occurs without restartable ECP, and this way ESP-IDF builds, where restartable ECP is the Kconfig flag `CONFIG_MBEDTLS_ECP_RESTARTABLE`, get the same behaviour.
- Add `Session::new_with_yield` to the blocking session: it takes a platform yield hook (`fn()`) invoked between restartable-ECC retry slices, so platforms with a scheduler can give up their time slice (e.g. `std::thread::yield_now`) instead of re-entering immediately. The existing `Session::new` delegates to it with a no-op hook.
- Add the safe `ecp::set_restartable_max_ops(u32)` wrapper. It must be called before creating `Tls` and panics if a `Tls` instance already exists, preventing safe callers from racing Mbed TLS's process-wide setting. The `ecp` module stays behind the `ecp-restartable` feature because Mbed TLS declares `mbedtls_ecp_set_max_ops` only under `MBEDTLS_ECP_RESTARTABLE`.
- Add `ClientSessionConfig::max_version`, an optional maximum TLS version. This crate always compiles TLS 1.3 in, and Mbed TLS 3.6 does not make the TLS 1.3 handshake crypto restartable, so a restartable client needs to be able to cap at TLS 1.2 — with TLS 1.2 the ECDSA signature and certificate verification are restartable whichever group the key exchange lands on. (The ordered key-exchange group allowlist that used to be in this PR is now its own PR: #176.)
- Document (on both session types' `new`) that a caller who wants to keep the stream across a failed creation can pass `&mut stream` — `Read`/`Write` are implemented for `&mut T`, and the borrow ends with the returned error.
- Add `Unreleased` changelog entries to both crates.

None of these APIs select a TLS version, cipher suite, or operation budget automatically.

## Example

Enable the feature together with the curve the application intends to use:

```toml
mbedtls-rs = { version = "0.3", features = ["embassy-time", "ecp-restartable", "curve-secp256r1"] }
```

Configure the process-wide budget before creating `Tls`, then apply any protocol policy explicitly to the client session:

```rust
use mbedtls_rs::{
    ecp::set_restartable_max_ops, ClientSessionConfig, Session, SessionConfig,
    Tls, TlsVersion,
};

// Configure once during initialization, before any Tls instance exists.
set_restartable_max_ops(100);
let tls = Tls::new(rng)?;

let config = SessionConfig::Client(ClientSessionConfig {
    max_version: Some(TlsVersion::Tls1_2),
    ..ClientSessionConfig::new()
});

let mut session = Session::new(tls.reference(), stream, &config)?;
session.connect().await?;
```

The TLS 1.2 cap above is application policy, not a default introduced by the feature. Mbed TLS documents that the operation budget is ignored by Curve25519, so an application that also wants to keep the key exchange on a short-Weierstrass curve can combine this with the group allowlist from #176.

## Compatibility and limitations

- Feature-disabled builds retain the previous generated configuration and session result handling.
- Blocking sessions run an in-progress restartable operation to completion by retrying, invoking the `new_with_yield` platform hook between slices (a no-op for sessions built with `Session::new`); only async sessions turn it into a cooperative yield.
- `ClientSessionConfig::max_version` defaults to `None`, preserving Mbed TLS's default maximum.
- `Session::new` remains source- and behavior-compatible. Adding public fields to the all-public `ClientSessionConfig` is a breaking change for struct literals built without `..`, so the release containing this change should bump the minor version (0.3, given 0.2.0 is now released).
- Enabling restartable ECC changes Mbed TLS C context layouts, so matching libraries and bindings must be built together.
- Mbed TLS 3.6 does not currently make the TLS 1.3 certificate-verification path restartable.

## Validation

- `cargo fmt --all --check`
- `cargo test --features use-gcc`
- `cargo test --features use-gcc,ecp-restartable`
- `cargo clippy -p mbedtls-rs --all-targets --features use-gcc,ecp-restartable -- -D warnings`
- `cargo check -p mbedtls-rs --features defmt,use-gcc`
- Source-build checks for the restartable configuration and a bare-metal RISC-V target

(The commands above were run with `use-gcc` because the development machine has no clang C compiler installed; environments with clang can drop it.)

Tests cover generated configuration and prebuilt invalidation, cooperative pending/self-wake/resume behavior, unchanged normal result paths, and `&mut`-stream recovery (a failed creation — an over-long server name — followed by a completed handshake over the same, still-owned stream). In addition, real loopback TLS handshakes run against committed self-signed fixtures: an async client/server pair over an in-memory duplex pipe and a blocking pair over a threaded pipe, each also exercised with `ecp-restartable`, an operation budget of 1 and a TLS 1.2 cap — the async variant asserts that cooperative self-wakes actually occurred, and the blocking variant asserts the handshake runs to completion and that the `new_with_yield` hook actually fires between slices.

## Structure

The branch is six commits on top of current `main` — the sys feature and configuration mapping, the session-side restartable handling, the client maximum TLS version plus the loopback handshake tests, and three review-feedback commits: the blocking session's `new_with_yield` hook, the by-`&mut` stream recovery documentation and test, and the unconditional in-progress handling. The key-exchange group allowlist has been split out into #176 per review; the two PRs are independent and can merge in either order (they share only the new `Unreleased` changelog heading, so whichever lands second gets a one-line rebase).

## Contribution terms

This contribution is submitted under the repository's existing dual MIT OR Apache-2.0 licensing terms, without additional terms or conditions.


mbedtls-rs/src/session/blocking.rs
```diff
@@ -324,6 +325,17 @@ where
             );
         }
 
+        // A blocking call cannot yield, so immediately re-enter a restartable ECC
```
ivmarkov: It actually can, but is platform/OS-specific.

For example, on OSes that do support threads and blocking IO, you might call `sleep(0)` which gives up the remaining time of your current time-slice to the scheduler.

But because yielding is platform specific, we have to make it flexible by introducing an extra parameter when constructing the blocking session, something like `yield: fn()` where you should call `yield()` just before the `continue;` on line 334.

Pass this to your LLM for assessment. In the meantime I'll review the rest today latest tmr.

But in general I like the idea of this PR!

mbedtls-rs/src/session/asynch.rs
```diff
@@ -67,9 +67,23 @@ where
...
+        Self::new_recoverable(tls, stream, config).map_err(|(_stream, error)| error)
+    }
+
+    /// Create a session, returning ownership of the stream if TLS state creation fails.
+    pub fn new_recoverable(
+        tls: TlsReference<'a>,
+        stream: T,
+        config: &SessionConfig<'a>,
+    ) -> Result<Self, (T, SessionError)> {
+        let state = match SessionState::new(config) {
```
ivmarkov: Can you check again whether - indeed - `SessionState::new` can fail with `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS` and if yes, explain where exactly in MbedTLS?

mbedtls-rs/src/session/blocking.rs
scottmcnab: How do you think this should work ergonomically? A couple of ideas:

1. Option<fn()> param:
```
let mut session = Session::new(
    tls.reference(),
    stream,
    &config,
    Some(|| std::thread::yield_now()),
)?;
                                              
// bare metal / no scheduler:
let mut session = Session::new(
    tls.reference(), stream, &config, None,
)?;
```

2. Required fn() param:
```
let mut session = Session::new(
    tls.reference(),
    stream,
    &config,
    || std::thread::yield_now(),
)?;
                                              
// bare metal / no scheduler:
let mut session = Session::new(
    tls.reference(), stream, &config, || {},
)?;
```

3. Feature-gated setter:
```
let mut session = Session::new(
    tls.reference(), stream, &config,
)?;
#[cfg(feature = "ecp-restartable")]
session.set_restartable_yield(|| {
     std::thread::yield_now()
 });
```

ivmarkov: Option 2 would be good enough, but I suggest - just like you have extra ctor fns for the async Session, maybe do the same for the blocking Session?

I.e. introduce `Session::new_with_yield` which takes this new `fn()` parameter.
Ant let the existing `Session::new` delegate to the new `Session::new_with_yield` with a `|| {}` impl?

mbedtls-rs/src/session/asynch.rs
scottmcnab: Checked again - no, `SessionState::new` cannot fail with `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS`, and `new_recoverable` isn't motivated by that error.

In Mbed TLS 3.6 that error is generated in exactly six places, all of them handshake steps converting `MBEDTLS_ERR_ECP_IN_PROGRESS` from a restartable ECC call:

- `library/ssl_tls12_client.c` - five sites in the TLS 1.2 client handshake (parse server key exchange, write client key exchange, certificate verify), each mapping `MBEDTLS_ERR_ECP_IN_PROGRESS` from restartable ECDH/ECDSA;
- `library/ssl_tls.c` - `mbedtls_ssl_verify_certificate`, mapping `MBEDTLS_ERR_ECP_IN_PROGRESS` from `mbedtls_x509_crt_verify_restartable`.

All of these are reachable only through handshake stepping (`mbedtls_ssl_handshake` / `mbedtls_ssl_read` / `mbedtls_ssl_write`). `SessionState::new` only calls `psa_crypto_init`, `mbedtls_ssl_config_defaults`, the `mbedtls_ssl_conf_*` setters (which store pointers - no EC arithmetic), `mbedtls_ssl_setup` (buffer allocation), and `mbedtls_ssl_set_hostname`, so none of the restartable paths can be hit there.

The actual motivation for `new_recoverable` is stream recovery: `Session::new` consumes the stream (usually a connected socket) and drops it when construction fails - allocation failure of the boxed SSL/config/DRBG contexts or of `mbedtls_ssl_setup`, or invalid configuration such as an empty group allowlist. `new_recoverable` returns the stream alongside the error so the application can close it gracefully or retry.

mbedtls-rs/src/session/blocking.rs
scottmcnab: Done - implemented as suggested in the latest commit: the blocking session gains `Session::new_with_yield(tls, stream, config, yield_fn: fn())`, with the hook invoked between the bounded slices of an in-progress restartable
operation, and the existing `Session::new` delegates to it with a `|| {}` no-op, so its signature is unchanged. The blocking restartable loopback test now installs a counting hook and asserts it actually fires during a budget-1
handshake.

mbedtls-rs/src/session/asynch.rs
ivmarkov: My point is the two error cases you outlined - allocation failure and invalid conf are typically non-recoverable. The latter error is also a programming bug.

I wonder - wouldn't just passing the stream by `&mut` in case the user wants to recover be good enough? Otherwise we need to introduce new_recoverable in the blocking session case as well, and combining that with new_with_yield becomes a bit too unwieldly.

scottmcnab: Agreed. Passing `&mut` is much better, and it needs no new API at all. Since `embedded-io` / `embedded-io-async` implement `Read` and `Write` for `&mut T`, a caller that wants the stream back passes `&mut stream` to the existing
`new`: on failure the borrow ends with the returned error and the stream is still theirs; on success the session just borrows it for its lifetime. That covers the case I actually cared about (getting an embedded socket back to reset it after an allocation failure), works identically for the blocking session, and doesn't interact with `new_with_yield` at all.

I've dropped `new_recoverable` from the PR in the latest commit. Both constructors now document the `&mut` pattern, and a loopback test exercises it against the real constructor: a failed creation (empty group allowlist) followed by a completed handshake over the same, still-owned stream. I also ported our firmware's two call sites to the pattern to confirm it works in practice.

mbedtls-rs/src/session/asynch.rs
```diff
@@ -981,6 +987,12 @@ where
                 mbedtls_ssl_set_bio(ssl_context, core::ptr::null_mut(), None, None, None);
             }
 
+            #[cfg(feature = "ecp-restartable")]
```
ivmarkov, line 990: I suggest we remove `#[cfg(feature = "ecp-restartable")]` and just have this code unconditionally. If the restarable support is not enabled, we'll never get `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS` so the code will not be triggered ayway.

Why removing the gate? 
Because `mbedtls-rs-sys` has a special mode of compilation on ESP-IDF targets (`*-espidf`), where it is just a thin Rust bindgen wrapper over the MbedTLS in ESP-IDF. 

In this mode of operation, whether MbedTLS is in restartable mode or not is controlled by a Kconfig flag - `CONFIG_MBEDTLS_ECP_RESTARTABLE`. 

While this is visible via `#[cfg(esp_idf_mbedtls_ecp_restartable)]` it is just easier to have all code handling `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS` errors unconditionally always compiled in, as it is not that big.

mbedtls-rs/src/session/blocking.rs
```diff
@@ -35,6 +35,12 @@ where
     /// `read`/`write`/handshake loops, which would otherwise only see an opaque
     /// MbedTLS code.
     last_io_err: Option<ErrorKind>,
+    /// Platform yield hook invoked between restartable-ECC retry slices.
+    ///
+    /// Only read when the `ecp-restartable` feature is enabled; stored
+    /// unconditionally so `new` and `new_with_yield` share one layout.
+    #[cfg_attr(not(feature = "ecp-restartable"), allow(dead_code))]
```
ivmarkov, line 42: Also here remove the cfg.

mbedtls-rs/src/session/blocking.rs
```diff
@@ -324,6 +364,19 @@ where
...
+        #[cfg(feature = "ecp-restartable")]
+        let result = loop {
+            let result = f(&mut self.state.ssl_context);
+            if result == MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS {
+                (self.yield_fn)();
+                continue;
+            }
+            break result;
+        };
+        #[cfg(not(feature = "ecp-restartable"))]
```
ivmarkov, line 379: Also here remove the cfg.

mbedtls-rs/src/session/blocking.rs
```diff
@@ -324,6 +364,19 @@ where
             );
         }
 
+        // Re-enter a restartable ECC operation in progress until it completes,
+        // invoking the platform yield hook between the bounded slices; only the
+        // async session turns it into a cooperative `Poll::Pending`.
+        #[cfg(feature = "ecp-restartable")]
```
ivmarkov, line 370: Also here remove the cfg.

mbedtls-rs/src/lib.rs
```diff
@@ -31,6 +31,8 @@ pub use session::*;
 pub(crate) mod fmt; // MUST be the first so that the other modules can see it
 
 mod cert;
+#[cfg(feature = "ecp-restartable")]
```
ivmarkov, line 34: Also here remove the cfg? That is, if `mbedtls_ecp_set_max_ops` is defined regardless of whether MbedTLS is compiled with or without restartable ECP support. Otherwise keep the flag.

mbedtls-rs/src/session.rs
```diff
@@ -309,6 +364,31 @@ impl<'a> Drop for ALPNArray<'a> {
     }
 }
 
+/// Owned, zero-terminated group-ID array allocated through Mbed TLS.
```
ivmarkov, line 367: Would you submit the `TlsGroup` support as a separate PR? I don't think this is related _at all_ to the restartable feature, so ideally, it should be a separate PR. Or am I missing something?

ivmarkov (changes requested):
I think we are almost there. Just two requests:
- Make the restart logic in `mbedtls-rs` unconditional, for the reason described in one of the review comments
- Remove the TlsGroup stuff from this PR and submit separately. I think it is unrelated to the restartable stuff anyway?

mbedtls-rs/src/session.rs
scottmcnab: Sorry yes thats my bad! This branch contained all the changes I was using for my project. I'll split it out into a separate PR.

mbedtls-rs/src/session/asynch.rs
scottmcnab: The `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS` handling in both sessions is no longer behind `ecp-restartable` (the async `Pending` + self-wake, the blocking retry loop and the `yield_fn` field). The `pub mod ecp` gate stays: Mbed TLS declares `mbedtls_ecp_set_max_ops` only under `MBEDTLS_ECP_RESTARTABLE` (`ecp.h`, the `#if defined(MBEDTLS_ECP_RESTARTABLE)` block around line 457–530 in 3.6), so it can't be compiled unconditionally. That's its own commit on the branch so the change is easy to see.

mbedtls-rs/src/session.rs
scottmcnab: Yes the group allowlist isn't part of the restartable machinery - the only link is Mbed TLS ignoring the operation budget for Curve25519, which is a reason to *want* an allowlist, not a dependency. It's now its own PR, #176, independent of this one (either order merges - they only share the new `Unreleased` changelog heading, and I can rebase whichever merges second).

One thing I kept here, in case you'd rather it moved too: `ClientSessionConfig::max_version`. This crate's dependency floor always compiles TLS 1.3 in, so every handshake negotiates 1.3 unless the client caps it - and Mbed TLS 3.6's TLS 1.3 handshake crypto isn't restartable (`MBEDTLS_SSL_ECP_RESTARTABLE_ENABLED` requires TLS 1.2 + ECDHE-ECDSA). Without a way to cap at 1.2, `ecp-restartable` can never actually yield through this crate's API, and the loopback tests that assert real cooperative yields rely on it. So I treated the version cap as the one prerequisite of the feature and kept it in this PR (a ~10-line addition). 

Happy to move it to #176 instead if you prefer.

PR #176, opened 2026-09-06
Add a client key-exchange group allowlist

scottmcnab:
## Summary

This adds an optional, ordered key-exchange group allowlist to `ClientSessionConfig`, split out of the original #170 pull-request.

`key_exchange_groups: Option<&'static [TlsGroup]>` lets an application state which named groups the client offers, in preference order. The IANA group IDs are copied into zero-terminated storage owned by the session state - as `mbedtls_ssl_conf_groups` requires the array to outlive the configuration - and handed to Mbed TLS during session creation. `None` (the default) preserves Mbed TLS's own group defaults; an empty allowlist is rejected at session creation with `MBEDTLS_ERR_SSL_BAD_INPUT_DATA`.

`TlsGroup` is a `#[repr(u16)]` enum over the `MBEDTLS_SSL_IANA_TLS_GROUP_*` constants (the NIST, Brainpool, Montgomery and FFDHE groups Mbed TLS knows about).

## Motivation

Group selection is application policy: pinning a deployment to the curves its peers and hardware accelerators actually support, or excluding a group for compliance reasons. It also pairs naturally with #170 - Mbed TLS documents that the restartable-ECC operation budget is ignored by Curve25519/448, so an application that wants bounded ECC slices needs to be able to keep the key exchange on a short-Weierstrass curve - but nothing here depends on that feature, which is why it is a separate PR.

## Changes

- Add `TlsGroup` and `ClientSessionConfig::key_exchange_groups`.
- Add an owned, zero-terminated group-ID array to the session state, allocated through Mbed TLS's allocator and dropped after the SSL context and configuration that reference it.
- Reject an empty allowlist at session creation.
- Add an `Unreleased` changelog entry.

Nothing selects a group automatically; `None` leaves Mbed TLS's defaults untouched.

## Example

```rust
use mbedtls_rs::{ClientSessionConfig, Session, SessionConfig, TlsGroup};

static GROUPS: &[TlsGroup] = &[TlsGroup::Secp256r1, TlsGroup::Secp384r1];

let config = SessionConfig::Client(ClientSessionConfig {
    key_exchange_groups: Some(GROUPS),
    ..ClientSessionConfig::new()
});

let mut session = Session::new(tls.reference(), stream, &config)?;
session.connect().await?;
```

## Compatibility

- `key_exchange_groups` defaults to `None`, so existing configurations negotiate exactly as before.
- Adding a public field to the all-public `ClientSessionConfig` is a breaking change for struct literals built without `..`, so the release containing this should bump the minor version (0.3, given 0.2.0 is released). The changelog entry is marked `(Breaking)` accordingly.
- No `mbedtls-rs-sys` changes: `mbedtls_ssl_conf_groups` and the IANA group constants are already in the bindings.

## Relationship to #170

The two PRs are independent and can merge in either order. They share only a byte-identical `[dev-dependencies]` block and identical test fixtures (which merge cleanly), plus the new `## [Unreleased]` changelog heading - whichever lands second will need a one-line changelog rebase, which I'll do promptly.

## Validation

- `cargo fmt --all --check`
- `cargo test --features use-gcc`
- `cargo clippy -p mbedtls-rs --all-targets --features use-gcc -- -D warnings`
- `cargo clippy --features use-gcc -- -D warnings`
- `cargo check -p mbedtls-rs --features defmt,use-gcc`

(`use-gcc` because the development machine has no clang C compiler; environments with clang can drop it.)

Tests exercise the real code paths rather than a model: the `TlsGroup` discriminants match the Mbed TLS IANA IDs; the default leaves the allowlist unset; an empty allowlist is rejected by the real constructor; a session with a two-group allowlist is created and dropped (allocation and drop order of the owned array); and a real blocking handshake with a P-256-only allowlist completes and echoes a payload over a localhost TCP loopback against committed self-signed ECDSA P-256 fixtures.

## Contribution terms

This contribution is submitted under the repository's existing dual MIT OR Apache-2.0 licensing terms, without additional terms or conditions.

### DIFF
--- a/mbedtls-rs/src/session.rs
+++ b/mbedtls-rs/src/session.rs
@@ -191,6 +191,8 @@ pub struct ClientSessionConfig<'a> {
     /// server (e.g. pinned or self-signed certificates served behind a
     /// different domain).
     pub skip_hostname_verification: bool,
+    /// Ordered key-exchange groups to offer. `None` keeps Mbed TLS's defaults.
+    pub key_exchange_groups: Option<&'static [TlsGroup]>,
 }
 
 impl<'a> Default for ClientSessionConfig<'a> {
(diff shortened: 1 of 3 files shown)


mbedtls-rs/src/session.rs
```diff
@@ -188,6 +188,8 @@ pub struct ClientSessionConfig<'a> {
     /// server (e.g. pinned or self-signed certificates served behind a
     /// different domain).
     pub skip_hostname_verification: bool,
+    /// Ordered key-exchange groups to offer. `None` keeps Mbed TLS's defaults.
+    pub key_exchange_groups: Option<&'static [TlsGroup]>,
```
ivmarkov: `&'a [TlsGroup]`

mbedtls-rs/src/session.rs
```diff
@@ -286,6 +320,16 @@ impl<'a> SessionConfig<'a> {
         }
     }
 
+    fn key_exchange_groups(&self) -> Option<&'static [TlsGroup]> {
```
ivmarkov: `&'a [TlsGroup]`

ivmarkov (commented):
Just one cosmetic fix: remove `'static` in favor of `'a` as the rest of the code.

Also need to resolve the conflicts, now that the restartable PR is merged.

PR #180, opened 2026-09-09
Option to fetch the negotiated TLS version

### LINKED ISSUE #167: Feature request: max TLS version cap + verify negotiated version after handshake
I'm using mbedtls-rs in a small embedded Rust project (not super experienced with Rust, especially not with FFI stuff, so bear with me if I get some of this wrong).

A couple of the devices I talk to have a known bug in their TLS 1.3 handling that causes corrupted transfers. Forcing TLS 1.2 for those specific connections apparently fixes it, but I also want to confirm afterward that 1.2 was actually used, since I don't want to silently trust an unverified connection.

It looks like `ClientSessionConfig` has `min_version`, but since both 1.2 and 1.3 are enabled by default and technically supported by these devices, setting a minimum wouldn't stop it from negotiating 1.3 anyway. I couldn't find a way to cap the max version, or any way to check the negotiated version on `Session` after connecting (only `tls_alpn()`/`tls_verification_details()` seem to exist for that kind of thing).

I poked around mbedtls's github a bit to see if this was possible there—well over my head—but saw they dealt with a [backwards compatibility breakage](https://github.com/Mbed-TLS/mbedtls/issues/9223) a few years ago that required the ability to cap TLS at 1.2, and gave me a clue with `mbedtls_ssl_conf_max_tls_version`, and i saw `mbedtls_ssl_get_version()` which returns a string value (not great, but useful).

It's not clear to me what would be usable on the rust side, and I don't want to guess further, but would it be possible to implement in mbedtls-rs:

1. A max_version field on `ClientSessionConfig` alongside min_version?
2. Some way to check the negotiated TLS version on `Session`?

Thanks for anything you could offer, and thank you for the crate!

ivmarkov:
The one thing remaining to close #167, and given how simple it is, why not?


bjoernQ (approved):
👍

PR #184, opened 2026-09-15
Raise the esp-idf-sys dep to 0.38

ivmarkov:
This should've been done _before_ the new 0.3.0 release form yesterday. :(

Since the "compile on top of ESP-IDF" branch is only used by `esp-idf-matter`, I suggest we release 0.3.1 and then yank 0.3.0.

The alternative of re-releasing `rs-matter`, `rs-matter-stack` and `openthread` one day later is just not worth it.


bjoernQ (approved):
👍

